PR #17, opened 2020-09-14
Swap 1042 dev environment

peter-asztalos:


martin-trajanovski:
@peter-asztalos I see tests are failing is it because of the changes in the docker-compose file or we should retry them?

peter-asztalos:
@martin-trajanovski this shouldn't affect the tests, the compose file has to be referenced directly to cause any effect (`docker-compose -f docker-compose.dev.yml ...`)

peter-asztalos:
@martin-trajanovski I think the previous failure was because the PR fixing tests wasn't merged, but it looks good now

martin-trajanovski (approved):

PR #36, opened 2020-09-22
Pull requests combined

jekabs-karklins:


jekabs-karklins:
@martin-trajanovski This is combined pullrequest of all 5 below pullrequest, when all changes are combined then E2E tests are passing.

PR #37, opened 2020-09-24
SWAP-1051 #resolved

martin-trajanovski:


martin-trajanovski:
This is probably failing because of the different name on the backend branch. We can first merge the backend branch into develop and it should pickup the latest develop and pass.

jekabs-karklins (approved):
Looks good

PR #39, opened 2020-09-28
Swap 1061 as a user i want to clone a sample

jekabs-karklins:


peter-asztalos (approved):

src/generated/sdk.ts
```diff
@@ -10,11 +10,15 @@ export type Scalars = {
   Boolean: boolean;
   Int: number;
   Float: number;
+  _Any: any;
```
martin-trajanovski, line 13: @jekabs-karklins I am not sure why we have/need this `_Any` scalar.

martin-trajanovski (approved):

src/generated/sdk.ts
peter-asztalos: 99% percent sure it is from Apollo federation, as we use it for generate the schema, even if we don't use the gateway or other features.

martin-trajanovski: Thanks good to know! 🚀

PR #41, opened 2020-10-06
Add Windows support

mr-field:
Part of UserOfficeProject/stfc-user-office-project#8

On Windows, specifying environment vars in package.json results in a build error so this PR moves `GENERATE_SOURCEMAP` to the `.env` file. It also adds a Windows-based Dockerfile.

The Dockerfile build stage it is identical to the existing one, but with a parameterised `FROM` statement that can take different Windows versions (see [here](https://docs.microsoft.com/en-us/virtualization/windowscontainers/manage-containers/hyperv-container) and [here](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility) for why we need separate images for each Windows version). NodeJS does not provide official Docker images for Windows, so the Dockerfile is based on our own NodeJS port hosted on an internal registry.

The Dockerfile run stage has been changed to run the frontend in IIS instead of Nginx, as Nginx does not have a stable Windows version.

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,2 @@
 REACT_APP_ORCID_REDIRECT=https://orcid.org/oauth/authorize?client_id=APP-IUN046ZZDWW0L6ZT&response_type=code&scope=/activities/update%20/read-limited&redirect_uri=https://useroffice.ess.eu/SignUp/
+GENERATE_SOURCEMAP=false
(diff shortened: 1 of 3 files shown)


jekabs-karklins (approved):

mr-field:
I've opened https://github.com/UserOfficeProject/stfc-user-office-project/issues/43 to look at alternatives for where the NodeJS base image is hosted.

PR #42, opened 2020-10-07
Swap 1062 as a user officer i want to have all steps editable

jekabs-karklins:


peter-asztalos (approved):

src/components/proposal/SampleDeclarationContainer.tsx
```diff
@@ -1,26 +1,27 @@
-import { Typography } from '@material-ui/core';
 /* eslint-disable @typescript-eslint/no-use-before-define */
+import { Typography } from '@material-ui/core';
```
martin-trajanovski, line 2: You can import this as default from @material-ui/core/Typography. I think it is the best to keep that practice.

martin-trajanovski (approved):
It looks good to me! 💯 Keep up the good work! 🔥

src/components/proposal/SampleDeclarationContainer.tsx
peter-asztalos: Actually I made a test on the scheduler side, using both (webpack should support tree shaking so it will only include the code you actually use), and the prod build was few kB smaller when I used `import {}  from '@material-ui/core';` than importing them directly.  But there may be differences in the react-scripts versions, maybe worth testing.

martin-trajanovski: When I was testing it the build and starting dev environment time was dropping up to 10s when I include them directly as defaults. Not sure but it was mentioned somewhere how to improve the build time in their react-scripts issues.

martin-trajanovski: Maybe we should try to update everything and do some testing again to see how to improve even more.

peter-asztalos: Maybe it affects build time, I didn't test that, only the build size.

PR #44, opened 2020-10-12
Swap 1052 attach workflow to a call

martin-trajanovski:


martin-trajanovski:
There are many changes but it is mainly because I removed all `cy.wait()` and introduced new commands like `cy.notification({ variant: 'success', text: 'Updated successfully' })`. Also I did some cleanup and tried to re-use the `useDataApiWithFeedback` hook created by @jekabs-karklins. It is very useful because it leads to much cleaner and shorter code but in the same time firing notifications is centralized at one place and you have better control.

martin-trajanovski:
This is how removing all the `cy.wait()` commands resulted:

Screenshot_15
Screenshot_16

peter-asztalos (approved):

PR #45, opened 2020-10-15
SWAP-1112 Scheduler Auth

peter-asztalos:


src/generated/sdk.ts
```diff
@@ -6683,6 +6683,14 @@ export function getSdk(client: GraphQLClient, withWrapper: SdkFunctionWrapper =
...
     verifyEmail(variables: VerifyEmailMutationVariables): Promise<VerifyEmailMutation> {
       return withWrapper(() => client.request<VerifyEmailMutation>(print(VerifyEmailDocument), variables));
+    },
+    checkToken(variables: {token: any}): Promise<any> {
+      return withWrapper(() => client.request<any>(print(gql`
+            query checkToken($token: String!) {
+        checkToken(token: $token) {
+          isValid
+        }
+      }`), variables))
```
peter-asztalos, line 6693: I'm waiting for a https://github.com/UserOfficeProject/user-office-frontend/pull/44 then this will be properly generated, currently the generate command fails

.env
```diff
@@ -1,2 +1,5 @@
 REACT_APP_ORCID_REDIRECT=https://orcid.org/oauth/authorize?client_id=APP-IUN046ZZDWW0L6ZT&response_type=code&scope=/activities/update%20/read-limited&redirect_uri=https://useroffice.ess.eu/SignUp/
-GENERATE_SOURCEMAP=false
\ No newline at end of file
+GENERATE_SOURCEMAP=false
+
+# in dev
+REACT_APP_AUTH_ALLOWED_HOSTNAMES=localhost
```
peter-asztalos, line 5: This will probably change after I figure out how to do this properly in different environments

src/generated/sdk.ts
martin-trajanovski: Merged it! 🦾

src/components/App.tsx
```diff
@@ -66,7 +67,9 @@ class App extends React.Component {
     try {
       error = JSON.parse(error);
     } catch (e) {}
```
martin-trajanovski, line 69: What we do here if it cannot be parsed?

peter-asztalos: I didn't write this part, but I assume we try to parse first (maybe it can be a stringified json) and if that fails we use the original version.

Actually I made some testing and this part of the code doesn't work as expected. useUnauthorizedApi has a hook inside it (useSnackbar), and hooks cannot be used this way so it will throw, so this part needs some refactor / review in the future.

mr-field (approved):

martin-trajanovski (approved):

PR #47, opened 2020-10-19
Swap 1062 as a user officer i want to have all steps editable

jekabs-karklins:
## Description

This pull request resolves user story: As a user officer, I want to have all steps editable in the template editor.

## Motivation

Previously, the first step was reserved and could not be modified. 
For Proposals that contained Title, Abstract, PI, and Co-proposers. 
For samples that contained Sample Title.
Now the first step is just like any other step, which can have other questions added. This is useful for samples (and in the future all the remaining templates that are going to be built) because some templates are short and one step is sufficient.

## Fixes:

* Story: As a user officer I want to have all steps editable in the template editor.
* Task: If there is only one step in the questionary, don't show the progress bar
* Bugfix: If there is an exception happening on the frontend while reporting it, another exception happens, effectively masking the original error
* Improvement: https://github.com/UserOfficeProject/user-office-frontend/pull/42#discussion_r502230097

## Changes:

* Template editor
* Questionary

- [ ] Added new tests for samples
- [ ] Added new test for proposal. As a user officer, I want to be able to modify the proposal


src/components/call/CallGeneralInfo.tsx
```diff
@@ -54,17 +54,29 @@ const CallGeneralInfo: React.FC = () => {
...
+        name="proposalWorkflowId"
+        label="Proposal workflow"
+        loading={loadingProposalWorkflows}
+        noOptionsText="No proposal workflows"
+        items={
+          proposalWorkflows.length > 0
+            ? proposalWorkflowsWithInjectedSelectionRemoval
+            : []
+        }
+        InputProps={{ 'data-cy': 'call-workflow' }}
```
jekabs-karklins, line 78: @martin-trajanovski Change#1 I added data-cy to InputProps so they end up being on the dropdown

src/graphql/call/fragment.call.graphql
```diff
@@ -23,4 +24,28 @@ fragment call on Call {
       ...basicUserDetails
     }
   }
+  proposalWorkflow {
```
jekabs-karklins, line 27: @martin-trajanovski Change #1 I added proposalWorkflow on the call fragment. I was wondering is this too expensive to fetch? If it is, then I'll make a task to refactor a little.

martin-trajanovski: @jekabs-karklins I see that on the call we don't need `proposalWorkflowConnectionGroups`. What we need is just id, name and description I guess. I think that is just what I added.

src/components/call/CallGeneralInfo.tsx
martin-trajanovski: This is good I wasn't aware of adding props on FormikDropdown like this. 🟢

peter-asztalos (approved):

martin-trajanovski (commented):
I recommend that we create smaller pull requests so it will be much easier to review. This one is very hard to review and it takes a lot of time.

cypress/integration/samples.ts
```diff
@@ -71,33 +90,52 @@ context('Samples tests', () => {
...
     cy.get('[data-cy=title-input] input')
       .clear()
-      .type(sampleTitle);
+      .type(sampleTitle)
+      .should('have.value', sampleTitle);
 
-    cy.get('[role="presentation"] [data-cy=save-and-continue-button]')
-      .first()
+    cy.get('[data-cy=save-and-continue-button]')
+      .eq(1)
```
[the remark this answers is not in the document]
jekabs-karklins: They are the same because the same component is reused. Good point, I'll check if I can target it better.

martin-trajanovski (approved):

cypress/integration/samples.ts
```diff
@@ -71,33 +90,52 @@ context('Samples tests', () => {
...
       .click();
 
     cy.wait(4000);
```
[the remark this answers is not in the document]
jekabs-karklins: The wait seems to not be needed and I added better targeting
063806ffa6ace6e565295b28a0a02862b10f6c6a

PR #49, opened 2020-10-20
Swap 1118 proposal workflow next status events

martin-trajanovski:
## Description

This pull request resolves the task: Ability to add multiple events that are triggering the change to the next status on a workflow connection.

## Motivation

Previously, we were not able to add any event that triggers the next status change in the UI and it was hardcoded on some `DEFAULT_EVENT`. Now you can add multiple events and they will be combined with the `AND` operator later when we check for status change in the `WorkflowEngine`.

## Fixes:

* Task: Ability to add multiple events that are triggering the change to the next status on a workflow connection.

## Changes:

Frontend adjustments for task resolution

### DIFF
--- a/src/components/common/FormikUICustomMultipleSelect.tsx
+++ b/src/components/common/FormikUICustomMultipleSelect.tsx
@@ -26,7 +26,7 @@ const FormikUICustomMultipleSelect = ({
   availableOptions: string[];
   id: string;
   label: string;
-  width: string;
+  width?: string;
 }) => {
   const ITEM_HEIGHT = 48;
   const ITEM_PADDING_TOP = 8;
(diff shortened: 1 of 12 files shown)


peter-asztalos (approved):

src/components/common/FormikUICustomMultipleSelect.tsx
```diff
@@ -25,14 +26,15 @@ const FormikUICustomMultipleSelect = ({
...
+  width: string;
 }) => {
   const ITEM_HEIGHT = 48;
   const ITEM_PADDING_TOP = 8;
   const MenuProps = {
     PaperProps: {
       style: {
         maxHeight: ITEM_HEIGHT * 4.5 + ITEM_PADDING_TOP,
-        width: 250,
+        width: width || 250,
```
jekabs-karklins, line 37: Maybe the width in props should be optional since it has a default value

jekabs-karklins (approved):

simonfernandes (approved):

PR #54, opened 2020-10-21
Swap 1129 fetch proposal events from backend

martin-trajanovski:
## Description

This pull request resolves the task: Create an endpoint to fetch all proposal events from the backend.

## Motivation

Previously, the proposal events were hardcoded in the frontend and now we are keeping track of events at only one place(backend) and send them to the frontend so the user officer should be able to select from those when creating the workflow.

## Fixes:

* Task: Create an endpoint to fetch all proposal events from the backend.

## Changes:

Frontend adjustments for task resolution


martin-trajanovski:
This should be merged after https://github.com/UserOfficeProject/user-office-frontend/pull/49

peter-asztalos (approved):

src/components/common/FormikUICustomMultipleSelect.tsx
```diff
@@ -13,6 +13,7 @@ const FormikUICustomMultipleSelect = ({
   availableOptions,
   id,
   label,
+  width,
```
jekabs-karklins, line 16: Maybe param should be optional as it has the default value

jekabs-karklins (approved):

src/components/common/FormikUICustomMultipleSelect.tsx
martin-trajanovski: I think I solved this in the previous PR https://github.com/UserOfficeProject/user-office-frontend/pull/49. And the changes are not here because I haven't pulled from thereafter I solved that. That is why I mentioned that this should be merged after https://github.com/UserOfficeProject/user-office-frontend/pull/49

mr-field (approved):

PR #64, opened 2020-10-26
Swap 495 as a user i should only see the public status of my proposal

jekabs-karklins:
## Description

User should only be able to see the public status of the proposal;

## Motivation and Context

We do not want to show user in which stage the proposal is in, because that is irrelevant and confusing for the user.
User only needs to see if the proposal is in the draft, submitted, accepted,  rejected or reserved

## How Has This Been Tested

e2e testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-495
As a user I should only see if my proposal is in the draft, submitted, rejected, Reserve or accepted
- Remove column submitted,
- In status show DRAFT, SUBMITTED and (ACCEPTED/REJECTED)
- All other statuses between SUBMITTED and ACCEPTED/REJECTED should be simply displayed as SUBMITTED
- Fix that status is always shown as DRAFT

## Changes

Updating ProposalTable to show public status instead of status

Relevant tests for changes are done in the backend.

### DIFF
--- a/cypress/integration/settings.ts
+++ b/cypress/integration/settings.ts
@@ -350,7 +350,7 @@ context('Settings tests', () => {
 
       cy.get('.MuiTable-root tbody tr')
         .first()
-        .then(element => expect(element.text()).to.contain('DRAFT'));
+        .then(element => expect(element.text()).to.contain('draft'));
 
       cy.get('.MuiTable-root tbody tr')
         .first()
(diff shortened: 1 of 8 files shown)


ACLay (approved):

martin-trajanovski:
@jekabs-karklins can you please check why the tests are failing here?

peter-asztalos (approved):

PR #65, opened 2020-10-28
Swap 1131 proposal workflow and status engine

martin-trajanovski:
## Description

This is a change based on the backend workflow engine resolution, fixing some bugs and improving the e2e tests. 

## Motivation and Context

This change is part of the proposal workflow implementation.

## How Has This Been Tested

It is tested with e2e tests using cypress that covers the whole workflow feature.

## Fixes

It fixes SWAP-1131 Jira task and it is part of https://github.com/UserOfficeProject/ess-user-office-project/issues/4

## Changes

Added DRAFT status as default one to the workflows when they are created. Tested the workflow engine feature. Fixing some bugs.

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -23,6 +23,7 @@
 //
 // -- This is will overwrite an existing command --
 // Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+import faker from 'faker';
 import { GraphQLClient } from 'graphql-request';
 
 const resetDB = () => {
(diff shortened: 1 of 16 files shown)


peter-asztalos (approved):

cypress/support/commands.js
```diff
@@ -107,6 +108,23 @@ const finishedLoading = () => {
   cy.get('[role="progressbar"]').should('not.exist');
 };
 
+const createProposal = (proposalTitle = '', proposalAbstract = '') => {
+  const title = proposalTitle || faker.random.words(3);
```
jekabs-karklins, line 112: faker.random ofter returns words with punctuations and dash or space-separated words. This can cause failed test false positives so maybe we should favour faker.lorem instead

jekabs-karklins (approved):

cypress/support/commands.js
martin-trajanovski: Yes I was thinking about that as well. But we should change it everywhere not only here then.

PR #70, opened 2020-11-05
Swap 1150 refactor questionarycomponents to keep all functio

jekabs-karklins:
## Description

This is purely a refactoring task. It does not add nor removes functionality but standardizes how QuestionaryComponents are implemented

Every component in the backend now implements a Question interface and by adding it to the registry it integrates into the questionary system. 

Here is the added documentation for more details.
https://github.com/UserOfficeProject/user-office-backend/wiki/How-to-create-new-questionary-component

## Motivation and Context

Since we expect to add more different types of components and that being the main way to extend functionality, components should be treated as components hiding behind an interface that can be added or removed from the system.

## How Has This Been Tested

- [x] Unit testing
- [x] Manual testing

## Fixes

SWAP-1150

## Changes

Questionary engine
Questionary components


src/components/questionary/QuestionaryComponentRegistry.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const definition = componentMap.get(id);
+  if (!definition) {
+    throw new Error(`Definition for ${id} was not found`);
+  }
+
+  return definition;
+}
+
+// Convenience methods below
+export const getQuestionaryComponentDefinitions = () => registry.slice(0);
```
peter-asztalos, line 73: Is there a reason for creating a shallow copy? The filter/map in `src/components/template/QuestionPicker.tsx` already creates a new array. If the goal is to have an array that can be freely altered, we should make a deep copy. If we want to have access to the list which can be freely filtered, used for mapping, using `ReadOnly` type will make it less error-prone I think.

peter-asztalos (approved):

src/components/questionary/QuestionaryComponentRegistry.tsx
jekabs-karklins: Good idea, also froze the object and did the same for the backend.

peter-asztalos: The `immer` package has a `Immutable<T>`  type, in the future may be worth using as it is usable for deep marking [like this](https://www.typescriptlang.org/play?ts=3.9.7#code/PQKhCgAIUhhB7AtgUwM6QGYCcmQJKIpZQjDgAuAngA7KQCC5SAlgMYDyARgFbKvmQAvOACQAH0gAxAK4A7fs3izREgLIBDagB51sygBpIuygD4VkAOrJ1Aaw3bjh42fGQAysnI69LiVdseXs7mAAo4iMyoyN6m5gAi6uTI5gBKyADmAKIAHtTmAELw8AA21squAHLSiJzIxK5u5FjMsung4Mi58FgCVLT4hNLk6pylWgAqJkKQ45CdSbIAJuiMLBw8fOSiAPwzogBcM3PZC8uQ8Bv8O5AA3ljWi0rFlJAA2gDSkC2QNsiU8BgZgBdQ4ERBDEZjcYfIEmAC+Bz2wGRKNRaPRGIxInarCUqAE6lQi2mrygkHJAEYAEwAZn0ZPJN0Ji0O1JpcPp5PJAHJmdz6UCANw4vG9NACQSQAAUAEpQYNhqNon1kACjESpoIpszheBcbJ8ZB7oDJUl8bLdcbXgAGIHTNnCq0UoEAOmZ9tpjuQGFeVLtkt5RO5gqAA) so it needs less manual typing.

jekabs-karklins:
Please look though, there are seemingly a lot of changes but most of it is just moving the same code do a different location :)

src/components/questionary/questionaryComponents/Embellishment/EmbellishmentDefinition.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { QuestionaryComponentDefinition } from '../../QuestionaryComponentRegistry';
+import { QuestionaryComponentEmbellishment } from './QuestionaryComponentEmbellishment';
+import { QuestionEmbellismentForm } from './QuestionEmbellismentForm';
+import { QuestionTemplateRelationEmbellismentForm } from './QuestionTemplateRelationEmbellismentForm';
+
+export const embellishmentDefinition: QuestionaryComponentDefinition = {
+  dataType: DataType.EMBELLISHMENT,
+  name: 'Embellishment',
+  questionaryComponent: QuestionaryComponentEmbellishment,
+  questionForm: () => QuestionEmbellismentForm,
```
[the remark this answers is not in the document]
jekabs-karklins: Fixed. Thanks.

jekabs-karklins:
@RasmiaKulan thanks for the "eagle eye", this has been resolved now.

PR #71, opened 2020-11-05
SWAP-1205 Fix scheduler triggered login

peter-asztalos:
## Description

This PR fixes an invalid app state which is caused by login triggered by the scheduler.

## Motivation and Context

A login triggered by the scheduler results in an invalid localStorage state which makes the frontend think the user is not logged in, although there is a token in the localStorage.

## How Has This Been Tested

existing e2e not failing

## Fixes

https://jira.esss.lu.se/browse/SWAP-1205

## Changes

- After login, we set the first available role regardless of how many roles a user has, previously it was done in `App.tsx`.
- We also wait for an available role and token before moving on.

## Depends on

No dependencies

### DIFF
--- a/src/components/user/SignIn.tsx
+++ b/src/components/user/SignIn.tsx
@@ -78,7 +78,7 @@ export default function SignInSide() {
   const classes = useStyles();
   const [failedLogin, setFailed] = useState(false);
   const [errorMessage, setErrorMessage] = useState('');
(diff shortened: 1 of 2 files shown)


martin-trajanovski (approved):

peter-asztalos:

jekabs-karklins (approved):

EdwardHaynes (approved):

PR #67, opened 2020-11-03
SWAP-1187 add support for sample pdf

peter-asztalos:
## Description

This PR adds support for sample PDF download.

## Motivation and Context

The goal is to support the sample PDF downloads.

## How Has This Been Tested

e2e tests were added.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1187

## Changes

- support for selecting samples
- support for downloading selected samples
- e2e tests updated
- some small changes/fixes

## Tests included/Docs Updated?

- [x] I have added tests to cover my changes.
- [ ] All relevant doc has been updated

PR #80, opened 2020-11-09
SWAP-1201 CI/CD Improvements

peter-asztalos:
## Description

This PR updates the Github action and Docker configurations to improve the overall speed of e2e testing.

## Motivation and Context

The previous solution didn't use any caching, was slow and inefficient.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1201

## Changes

- save intermediate layers after each build and use them if there were no changes
- docker ignore updated to provide a better caching experience
- build and store the cypress image in the Github package registry, this improves pulling time drastically
- mount the e2e tests as a volume, so we can use the built image
- use the pre-built and cached frontend image in testing and when pushing the new version
- some files removed and some parts move to the corresponding service's repo

## Depends on

https://github.com/UserOfficeProject/user-office-frontend/pull/67


peter-asztalos:
The changes include https://github.com/UserOfficeProject/user-office-frontend/pull/67 as it wasn't merged yet.

peter-asztalos:

martin-trajanovski (approved):

.github/workflows/build-push.yml
```diff
@@ -1,20 +1,51 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+# This is a basic workflow to help you get started with Actions
```
jekabs-karklins, line 1: The other comments for steps are nice, but maybe this one should be removed.

Dockerfile
```diff
@@ -1,13 +1,20 @@
 # Stage 0, "build-stage", based on Node.js, to build and compile the frontend
-FROM node:10.22.0-alpine as build-stage
+FROM node:12-alpine as build-stage
```
jekabs-karklins, line 2: We need to keep an eye on stability after the upgrade

jekabs-karklins (approved):

Dockerfile
peter-asztalos: We should be fine, Node is only used for building, in local we were already running Node 12 for development and backend.

vovsike (approved):

PR #81, opened 2020-11-10
Swap 1166 question order not saved

martin-trajanovski:

### DIFF
--- a/src/components/template/TemplateEditor.tsx
+++ b/src/components/template/TemplateEditor.tsx
@@ -25,6 +25,7 @@ import {
   getTopicById,
 } from 'models/QuestionaryFunctions';
 import { StyledPaper } from 'styles/StyledComponents';
+import { randomNumberBetween } from 'utils/Math';
 
 import QuestionEditor from './forms/QuestionEditor';
 import QuestionTemplateRelationEditor from './forms/QuestionTemplateRelationEditor';
@@ -93,7 +94,7 @@ export default function TemplateEditor() {
     },
   }))();
 
-  const getTopicListStyle = (isDraggingOver: any) => ({
+  const getTopicListStyle = (isDraggingOver: boolean) => ({
     background: isDraggingOver
       ? theme.palette.primary.light
       : theme.palette.grey[100],
@@ -101,11 +102,15 @@ export default function TemplateEditor() {
     display: 'flex',
   });
 
-  const randomNumberBetween = (min = 0, max = 1) => {
-    return Math.random() * (max - min) + min;
-  };
-
   const onDragEnd = (result: DropResult): void => {
+    if (
+      !result.destination ||
+      (result.destination.droppableId === result.source.droppableId &&
+        result.destination.index === result.source.index)
+    ) {
+      return;
+    }
+
     const isDraggingQuestion = result.type === 'field';
     const isDraggingTopic = result.type === 'topic';
 
@@ -113,10 +118,14 @@ export default function TemplateEditor() {
       const dragSource = result.source;
       const dragDestination = result.destination;
       const isDraggingFromQuestionDrawerToTopic =
-        dragSource.droppableId === 'questionPicker';
+        dragSource.droppableId === 'questionPicker' &&
+        dragDestination?.droppableId !== 'questionPicker';
       const isDraggingFromTopicToQuestionDrawer =
-        dragDestination?.droppableId === 'questionPicker';
-      const isReorderingInsideTopics = dragDestination && dragSource;
+        dragDestination?.droppableId === 'questionPicker' &&
+        dragSource.droppableId !== 'questionPicker';
+      const isReorderingInsideTopics =
+        dragDestination?.droppableId !== 'questionPicker' &&
+        dragSource.droppableId !== 'questionPicker';
 
       if (isDraggingFromQuestionDrawerToTopic) {
         const questionId =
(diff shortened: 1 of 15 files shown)


jekabs-karklins (approved):

peter-asztalos (approved):

src/utils/Math.ts
```diff
@@ -1,3 +1,7 @@
-export function clamp(input: number, lower: number, upper: number): number {
+export const clamp = (input: number, lower: number, upper: number): number => {
   return Math.min(Math.max(input, lower), upper);
-}
+};
+
+export const randomNumberBetween = (min = 0, max = 1): number => {
```
ACLay, line 5: This function is only called when getting the sorting position between two items, and it seems odd to be calling a random number generator for that. I feel like using an average would give better performance.

src/hooks/questionary/usePersistQuestionaryEditorModel.ts
```diff
@@ -224,31 +203,76 @@ export function usePersistQuestionaryEditorModel() {
...
+          executeAndMonitorCall(() =>
+            updateQuestionTopicRelation(state.templateId, questionRel)
+          );
           break;
-        case EventType.REORDER_TOPIC_REQUESTED:
-          const topicOrder = state.steps.map(step => step.topic.id);
-          executeAndMonitorCall(() => updateTopicOrder(topicOrder));
+        case EventType.REORDER_TOPIC_REQUESTED: {
+          const sourceIndex = action.payload.source.index;
+          const destinationIndex = action.payload.destination.index;
```
ACLay, line 236: These shorter variable names could be used again in later lines instead of the full `action.payload...`

ACLay (changes requested):

src/graphql/template/createQuestionTemplateRelation.graphql
```diff
@@ -2,7 +2,7 @@ mutation createQuestionTemplateRelation(
   $templateId: Int!
   $questionId: String!
   $topicId: Int!
-  $sortOrder: Int!
+  $sortOrder: Float!
```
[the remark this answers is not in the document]
peter-asztalos: Actually, these files are used in generating the `src/generated`, the generator can generate the TypeScript types using the GraphQL graphs on the backend, and using the files in `src/graphql` we tell the generator which queries/mutations we want to actually use, so it can generate the required types/calls and also check if our queries/mutations are actually typesafe from the Graphql point of view.
If you try to access some fields which don't exist, or you use wrong types in the parameters the generator will fail.

src/hooks/questionary/usePersistQuestionaryEditorModel.ts
```diff
@@ -224,31 +203,76 @@ export function usePersistQuestionaryEditorModel() {
...
-            );
+            destinationTopic = extendedTopic;
           }
+
+          const questionRelToChange =
+            destinationTopic?.fields[action.payload.destination.index];
+          const prevQuestion =
+            destinationTopic?.fields[action.payload.destination.index - 1];
+          const nextQuestion =
+            destinationTopic?.fields[action.payload.destination.index + 1];
```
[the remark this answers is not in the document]
martin-trajanovski: Yes this is handled with default values in the sort order calculations. For example if one of the values or both are `undefined` it will result with number between some defaults.

src/utils/Math.ts
martin-trajanovski: Maybe if I go with the solution of storing integers as we discussed on the backend part I wont need this. I am thinking of going back to integers and update all the later items in the list.

src/components/template/TemplateEditor.tsx
```diff
@@ -171,9 +171,29 @@ export default function TemplateEditor() {
...
       } else if (isReorderingInsideTopics) {
+        // TODO: Put this into a function because it is repeated ------
+        const topicId = dragDestination?.droppableId
+          ? +dragDestination.droppableId
+          : undefined;
+        const topic = getTopicById(state.steps, topicId as number);
+
+        const previousField =
+          topic.fields[(dragDestination?.index as number) - 1];
+        const nextField = topic.fields[dragDestination?.index as number];
```
peter-asztalos, line 182: I think we could check if `dragDestination` exists at the beginning then we can avoid the castings. That should make the compiler happy while making the code more safe. The same goes for the other variables used.

src/models/QuestionaryEditorModel.ts
```diff
@@ -90,14 +90,18 @@ export default function QuestionaryEditorModel(
               step.topic.id.toString() ===
               action.payload.destination.droppableId
             );
-          });
+          }) as TemplateStep;
```
peter-asztalos, line 93: Is it 100% the `draft.steps` contains the `droppableId`? Otherwise, I suggest checking `to` and `from`.

src/components/template/TemplateEditor.tsx
martin-trajanovski: I think I check this on line `107` but I should put the check on `dragDestination` instead of `result.destination` that will be better.

src/models/QuestionaryEditorModel.ts
martin-trajanovski: I am working on this right now and will change the approach to use integers with updating all later items. I will go trough this again. I better put this as WIP.

martin-trajanovski:
@ACLay I think now it is ready to be reviewed again with the requested changes. I went with the approach of updating `sort_order` field on all other items.

ACLay (changes requested):

ACLay (approved):

PR #93, opened 2020-11-16
Swap 1141 Adding interval component

jekabs-karklins:
## Description

Adding interval component 

## Motivation and Context

This merge request adds a new component called Interval Component. With it, the user can answer the question which requires to specify "min", "max" and "units".

Without this component, there would be a need for three questions that would not be convenient to create and fill out.

## How Has This Been Tested

e2e tests
manual testing

## Fixes

SWAP-1141 As a user I should be able to answer an interval question

## Changes

New Questionary Component

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/73


peter-asztalos (approved):

src/components/questionary/questionaryComponents/SampleDeclaration/QuestionSubtemplateForm.tsx
```diff
@@ -64,42 +61,6 @@ export const QuestionSubtemplateForm: FormComponent<Question> = props => {
...
-                      value={category.categoryId}
-                      key={category.categoryId}
-                    >
-                      {category.name}
-                    </MenuItem>
-                  );
-                })}
-              </Field>
-            </FormControl>
-
```
mr-field, line 102: I'm not sure why this block was deleted. Did it interfere with the interval questions?

mr-field (commented):
Looks good, just one question (maybe just me being thick)

src/components/questionary/questionaryComponents/SampleDeclaration/QuestionSubtemplateForm.tsx
jekabs-karklins: I'll move this change do a different pull-request, as it does not really make sense to be part of "Adding interval component"

mr-field (approved):

PR #97, opened 2020-11-18
Swap 1199 filtering based on statuses

martin-trajanovski:
## Description

Filtering the proposals by the statuses from the workflow.

## Motivation and Context

This is the solution for better following of the proposal workflow. If you as a Instrument Scientist need to see only proposals that are for feasibility review they are filtered by default on that.

## How Has This Been Tested

With e2e tests.

## Fixes

It fixes https://jira.esss.lu.se/browse/SWAP-1199

## Changes

Added filter by proposal status in the proposal filter bar and refactored some code and components to be more configurable.


peter-asztalos (approved):

jekabs-karklins (approved):

martin-trajanovski:
@EdwardHaynes can you please take a look at this one as well because I see you already reviewed the backend part?

EdwardHaynes (approved):

PR #101, opened 2020-11-24
Swap 1235 default proposal statuses

martin-trajanovski:
## Description

Adding more default proposal statuses and added `shortCode` field to be able to use name in a more human readable way. 

## Motivation and Context

This is solving the need for more default proposal statuses that are going to be used inside proposal workflows and need for more human readable name everywhere we use the proposal statuses.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1235

## Changes

Changes in the database to add new field and some refactoring to be able to use it.

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/86


src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -305,6 +303,11 @@ const ProposalTableOfficer: React.FC<ProposalTableOfficerProps> = ({
 
     const isError = !!assignmentsErrors.join('');
 
+    enqueueSnackbar('Proposal/s assigned to SEP', {
```
peter-asztalos, line 306: I think we need a different message when it fails

peter-asztalos (approved):

jekabs-karklins (approved):

PR #105, opened 2020-11-25
SWAP-1196 As a developer I want Frontend to use develop if respective branch does not exist

peter-asztalos:
## Description

This PR aims to fix falling actions where the feature branch no longer exists, but we still have the docker image, we do this by checking if both the docker image and the feature branch exists.

## Motivation and Context

"Currently, the build will fail if the docker image is found but the branch isn't
We want to fallback checking out develop in case the corresponding backend branch does not exist"

## How Has This Been Tested

- [x] Manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-1196

## Changes

- action script updated, added support to check both Docker Hub and Github for the feature branch


jekabs-karklins (approved):
Awesome

PR #113, opened 2020-11-26
Swap 1209 Bugfix. Allow users to input multiple line text on multi-line input

jekabs-karklins:
## Description

This is a bugfix. 
Allow users to input multiple line text on multiline input.

## Motivation and Context

The bug is a regression bug introduced when preventing the submission of the form when ENTER is pressed on a text input field. On multiline text inputs, however, pressing ENTER is acceptable and no default behavior should be prevented.

## How Has This Been Tested

Added new e2e test to prevent breaking the feature again.

## Fixes

SWAP-1209 - User can't enter multiple lines in multi-line input

## Changes

QuestionaryComponentTextInput

## Depends on

N/A


peter-asztalos (approved):

src/components/questionary/questionaryComponents/TextInput/QuestionaryComponentTextInput.tsx
```diff
@@ -1,7 +1,6 @@
 import makeStyles from '@material-ui/core/styles/makeStyles';
 import { getIn } from 'formik';
 import React, { ChangeEvent, KeyboardEvent, useEffect, useState } from 'react';
-import { Key } from 'ts-keycode-enum';
```
martin-trajanovski, line 4: Is there a specific reason why we stop using `ts-keycode-enum`? And I see that `event.keyCode` is deprecated and we should not use it anymore I guess. https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode I see there is one more place that we still use it in our frontend code (src\components\questionary\questionaryComponents\SampleBasis\QuestionaryComponentSampleBasis.tsx). It will be good if you change it as well and maybe remove the `ts-keycode-enum` package from package.json

martin-trajanovski (approved):

src/components/questionary/questionaryComponents/TextInput/QuestionaryComponentTextInput.tsx
jekabs-karklins: This has been removed in https://github.com/UserOfficeProject/user-office-frontend/pull/121/files#diff-50603e48c1941520a0287e3a8c9156d5666488424355f292727152034273b70dL9

martin-trajanovski: But is it removed from the `package.json` because now it is not in use or?

PR #114, opened 2020-11-27
Swap 1224 save table sort in url params

martin-trajanovski:
## Description

The changes are because of the need to keep the sorting of tables data by columns. Now when you try to sort by some column  the sort parameters are saved as url params and they are not lost if you reload or table gets re-rendered. Also added some tests for this feature and for switching roles inside not allowed routes.

## How Has This Been Tested

E2E tests are included.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1224

## Changes

Changes are mostly in the SuperMaterialTable component because it is used on most of the places where we need to show data in a table.


src/components/common/SuperMaterialTable.tsx
```diff
@@ -62,6 +79,17 @@ function SuperMaterialTable<Entry extends EntryID>({
     props.options.searchText = props.urlQueryParams.search || undefined;
   }
 
+  const { columns } = props;
```
peter-asztalos, line 82: I think we should be consistent, if already use destructuring for the column, we should use the same for e.g. `urlQueryParams`.

peter-asztalos (approved):

src/components/common/SuperMaterialTable.tsx
martin-trajanovski: Yes that is a good idea I will try to extract all the properties from props at once and then use them.

cypress/integration/proposal_administration.ts
```diff
@@ -137,4 +137,61 @@ context('Proposal administration tests', () => {
...
     cy.get('[placeholder="Search"]').should('have.value', 'test');
   });
+
+  it('Should be able to save table sort state in url', () => {
+    let officerProposalsTableAsTextBeforeSort = '';
+    let officerProposalsTableAsTextAfterSort = '';
+
+    cy.login('user');
+    cy.createProposal(
+      'Aaaaaaaaa test proposal title',
```
jekabs-karklins, line 147: I see no reason for multiple 'Aaaaaaaaa'. A typo?

jekabs-karklins (approved):

cypress/integration/proposal_administration.ts
martin-trajanovski: I have added multiple `Aaaaaaa` just because here I am testing the sorting by title if it is kept in the url prams after reloading. There is already one proposal with a random title and this one I wanted to be always on top if the sort order is ascending and always at the bottom if the sort order is descending. Maybe it looks a bit weird but didn't have any better idea how to do it 😅

Scott-James-Hurley (approved):

PR #118, opened 2020-11-30
SWAP-115 Trying to signup with taken username results in exception

peter-asztalos:
## Description

Show error messages sent by the backend and block additional submit while submitting

## How Has This Been Tested

- [x] e2e test
- [x] Manual 

## Fixes

https://jira.esss.lu.se/browse/SWAP-115

## Changes

- try to translate and show backend error messages
- block form submit while submitting

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/94


martin-trajanovski (approved):
This looks great but I guess we need some more tests that will cover some other cases for example where we do not fill in all the required fields and so on. Just to catch scenarios like this one where we think that some field is not required but the validation is wrong and if you do not provide it an error will occur.

PR #119, opened 2020-11-30
SWAP-982 #resolved and implement call partial validation

martin-trajanovski:
## Description

Create/Update call validation schemas updated and split and fit the step validation needs.

## Motivation and Context

The point is to have step validation on the frontend and full validation on the backend.

## How Has This Been Tested

With e2e tests on the frontend.

## Fixes

https://jira.esss.lu.se/browse/SWAP-982

## Depends on

https://github.com/UserOfficeProject/user-office-lib/pull/18
https://github.com/UserOfficeProject/user-office-backend/pull/95

### DIFF
--- a/src/components/call/CreateUpdateCall.tsx
+++ b/src/components/call/CreateUpdateCall.tsx
@@ -1,9 +1,8 @@
 import {
-  createCallValidationSchema,
-  updateCallValidationSchema,
+  createCallValidationSchemas,
+  updateCallValidationSchemas,
 } from '@esss-swap/duo-validation/lib/Call';
 import Button from '@material-ui/core/Button';
-import FormHelperText from '@material-ui/core/FormHelperText';
 import Step from '@material-ui/core/Step';
 import StepLabel from '@material-ui/core/StepLabel';
 import Stepper from '@material-ui/core/Stepper';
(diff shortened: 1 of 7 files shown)


peter-asztalos (approved):

src/components/call/CreateUpdateCall.tsx
```diff
@@ -106,15 +72,47 @@ const CreateUpdateCall: React.FC<CreateUpdateCallProps> = ({ call, close }) => {
...
+
+  const handleBack = (formProps: FormikProps<Call>) => {
+    formProps.setTouched({});
+
     setActiveStep(prevActiveStep => prevActiveStep - 1);
   };
 
-  const handleStep = (step: number) => () => {
+  const handleStep = (step: number, formProps: FormikProps<Call>) => () => {
+    formProps.setTouched({});
```
jekabs-karklins, line 114: What does the setTouched() do in this case?

jekabs-karklins (approved):

src/components/call/CreateUpdateCall.tsx
martin-trajanovski: It was used in some examples that I have seen in formik github repo. It is mainly because when you change from one to another step and try to check `formProps.isValid` is always initially `false` even if you have all the valid fields. This is like a trigger to just to change the state if everything is valid inside selected step.

PR #123, opened 2020-12-02
SWAP-1242 Upgrade Formik version to v2 in User Office Frontend

peter-asztalos:
## Description

This PR upgrades Fromik and related packages to a newer version and also fixes issues caused by the upgrade.

## How Has This Been Tested

- [x] e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1242

## Changes

- packages upgraded
- issues caused by upgrade fixed
- unnecessary code parts removed


src/components/SEP/Members/SEPMembers.tsx
```diff
@@ -234,13 +234,15 @@ const SEPMembers: React.FC<SEPMembersProps> = ({ sepId }) => {
         title={'SEP Secretary'}
         invitationUserRole={UserRole.SEP_SECRETARY}
       />
+      {/**
+       * Note: Do we actually need Formik here? It looks like it could be removed
+       *       and we can use regular React hooks and Mui form components
+       */}
```
martin-trajanovski, line 240: Maybe you are right there is no need for including formik here because we are not submitting anything. We are updating things one by one. I will check this.

peter-asztalos: I see, then we can resolve it in a different ticket/PR.

martin-trajanovski (approved):
Looks great! I like when we fix some things, update packages and remove more lines of code then we add.

jekabs-karklins:
@peter-asztalos  Please don't merge before I'll check

peter-asztalos:
> @peter-asztalos Please don't merge before I'll check

Okay, actually that was my plan.

src/components/questionary/QuestionaryStepView.tsx
```diff
@@ -99,7 +100,12 @@ export default function QuestionaryStepView(props: {
                             newValue: newValue,
                           },
                         });
-                        handleChange(evt);
+
+                        if (typeof evt === 'string') {
+                          setFieldValue(evt, newValue, true);
```
jekabs-karklins, line 105: Hi, Peter, I like this part where you using setFieldValue.
I believe we should use setFieldValue always and get rid of handleChange altogether. Just need to change the signature of onComplete. I will do this in a separate pullRequest, so looks good for now.

jekabs-karklins (approved):
Looks good

PR #124, opened 2020-12-03
Swap 1200 proposal basic information is not validated

jekabs-karklins:
## Description

This is a bugfix fixing a bug where the user could create a proposal without specifying title and abstract

## Motivation and Context

I added the YupValidation schema to ProposalBasis, and that fixed the input validation.
I also added a new function to the QuestionaryComponent interface called getYupInitialValue which is used to build up Yup initial values.

## How Has This Been Tested

- [x] E2E
- [x] Manual testing

## Fixes

SWAP-1200 proposal basic information is not validated

## Depends on

N/A


peter-asztalos (approved):

src/components/questionary/questionaryComponents/ProposalBasis/proposalBasisValidationSchema.ts
```diff
@@ -0,0 +1,25 @@
+import * as Yup from 'yup';
+
```
martin-trajanovski, line 2: Maybe this validation schema should be part of the validation package or what do you think?

martin-trajanovski (approved):

src/components/questionary/questionaryComponents/ProposalBasis/proposalBasisValidationSchema.ts
jekabs-karklins: Great idea, although this requires a bigger refactor what I should definitely do maybe in the next sprint. I created a task describing a task
https://jira.esss.lu.se/browse/SWAP-1277

PR #138, opened 2020-12-15
SWAP-1188 As a user officer I should be able to verify users

peter-asztalos:
## Description

This PR adds support to manually verify email and remove the placeholder flag from a user account.

## Motivation and Context

As it is now, it is not possible to do this manually as a User Officer.

## How Has This Been Tested

- [x] e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1188

## Changes

- API endpoints added for the manual updates
- UI updated

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/107

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,4 @@
 {
   "video": false,
-  "baseUrl": "http://127.0.0.1:3000"
+  "baseUrl": "http://127.0.0.1"
 }
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,4 @@
 {
   "video": false,
-  "baseUrl": "http://localhost"
+  "baseUrl": "http://127.0.0.1:3000"
 }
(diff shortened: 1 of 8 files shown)


cypress.json
```diff
@@ -1,4 +1,4 @@
 {
   "video": false,
-  "baseUrl": "http://localhost"
+  "baseUrl": "http://127.0.0.1"
```
peter-asztalos, line 3: This drastically reduces the test's runtime, looks like DNS resolving can be very slow when Cypress is open (on Windows at least ).
In the Docker environment, it seems to be fine as I had tests finishing without issues in docker headless but timing out on Windows with cypress open.

mr-field (approved):

martin-trajanovski (approved):

PR #139, opened 2020-12-15
Add external auth support

mr-field:
Part of UserOfficeProject/stfc-user-office-project#44

## Description
See initial design at UserOfficeProject/stfc-user-office-project#48

This PR adds a new endpoint that takes an authentication token generated by an external application. The token is passed to the new mutation introduced in the [backend PR](https://github.com/UserOfficeProject/user-office-backend/pull/108), and the user is either logged in or redirected to the external login page depending on the result of the muation.

All "user-management" endpoints are now enabled or disabled based on the value of the `REACT_APP_AUTH_TYPE` environment variable. The new endpoint is enabled and all other user-related endpoint are disabled only when this variable is set to `external`.

## Motivation and Context

This will allow STFC users to log into the proposals app with their existing account.

## How Has This Been Tested
This was tested manually on my local machine. Since the UserOfficeWebService is only accessible within the STFC network, it would be tricky to test this using e2e tests.

## Changes
A new React component was added to handle the external authentication requests. The available React routes are now conditional  on the `REACT_APP_AUTH_TYPE` environment variable.

## Depends on
https://github.com/UserOfficeProject/user-office-backend/pull/108


bolmsten:
You are adding variables such as REACT_APP_AUTH_TYPE, I think this will require a special build for STFC frontend container. If you would get that value from the backend you should be able to use the same container that is in docker hub

peter-asztalos (approved):

ACLay (approved):

PR #142, opened 2020-12-17
SWAP-1268 #resolved

martin-trajanovski:
## Description

This fixes the frontend bug related to multi-choice question dependencies.

## Motivation and Context

There was a bug when you try to add multi-choice question as dependency to some other questions. As well added some reusable commands for dragging and dropping draggable elements.

## How Has This Been Tested

E2E tests.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1268

## Changes

Changes mainly in the comparator for question dependencies and added reusable command for moving elements because that logic is used across multiple test files.


cypress/support/commands.js
```diff
@@ -126,6 +134,25 @@ const createProposal = (proposalTitle = '', proposalAbstract = '') => {
   cy.contains('Save and continue').click();
 };
 
+const dragElement = (element, moveArgs) => {
```
martin-trajanovski, line 137: Added new command so anytime someone needs to move element somewhere can use it. 🐉

peter-asztalos (approved):

jekabs-karklins (approved):

PR #143, opened 2020-12-17
SWAP-1289 User office frontend: upgrade Cypress to the latest version (v6.1.0)

peter-asztalos:


cypress/package.json
```diff
@@ -11,6 +11,6 @@
   },
   "scripts": {
     "cypress:open": "cypress open",
-    "cy:run": "cypress run --config defaultCommandTimeout=15000"
+    "cy:run": "cypress run --browser chrome --config defaultCommandTimeout=15000"
```
peter-asztalos, line 14: Currently, there is an issue with Electron, see https://github.com/cypress-io/cypress/issues/8865

martin-trajanovski: Is there any change required to be able to run on windows or?

jekabs-karklins (approved):

cypress/package.json
peter-asztalos: I think it should be fine, if you have issues with Electron throwing the mentioned error you can just use Edge or Chrome or ignore it, it's more relevant in CIs as it could block merging the PRs.

martin-trajanovski (approved):

PR #166, opened 2021-01-06
Swap 1273 Add confirmation box when copying sample

bolmsten:
## Description

When a user copies a sample a confirmation box will appear

## Motivation and Context

Requested by users at ESS


jekabs-karklins (approved):
Just FYI. There is also a HOC withConfirm(), which passes the confirm method to the wrapped component props.

Scott-James-Hurley (approved):

PR #178, opened 2021-01-11
SWAP-1230 show error if upload fails

jekabs-karklins:
## Description

If the file upload fails it should be handled gracefully in the frontend


peter-asztalos:
The PR has no changes.

jekabs-karklins:
Oh, so I guess this got merged then after all. Thanks @peter-asztalos

PR #181, opened 2021-01-13
SWAP-379 Make design consistent across questionary

peter-asztalos:
## Description

This PR tries to fix some design inconsistency in questionaries.

## Motivation and Context

There are some missing or inconsistently looking components used by the questionaries.

## How Has This Been Tested

- [x] existing e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-379

## Changes

- questionary components style refactored
- some small fixes
- some typos fixed

### DIFF
--- a/cypress/integration/samples.ts
+++ b/cypress/integration/samples.ts
@@ -93,6 +93,8 @@ context('Samples tests', () => {
       '[data-cy=sample-declaration-modal] [data-cy=save-and-continue-button]'
     ).click();
 
+    cy.finishedLoading();
+
     cy.get('[data-cy="questionnaires-list-item"]').should('have.length', 1);
 
     cy.get('[data-cy="clone"]').click();
--- a/cypress/integration/samples.ts
+++ b/cypress/integration/samples.ts
@@ -93,17 +93,17 @@ context('Samples tests', () => {
       '[data-cy=sample-declaration-modal] [data-cy=save-and-continue-button]'
     ).click();
 
-    cy.get('[data-cy="questionaries-list-item"]').should('have.length', 1);
+    cy.get('[data-cy="questionnaires-list-item"]').should('have.length', 1);
 
     cy.get('[data-cy="clone"]').click();
 
-    cy.get('[data-cy="questionaries-list-item"]').should('have.length', 2);
+    cy.get('[data-cy="questionnaires-list-item"]').should('have.length', 2);
 
     cy.get('[data-cy="delete"]')
       .eq(1)
       .click();
 
-    cy.get('[data-cy="questionaries-list-item"]').should('have.length', 1);
+    cy.get('[data-cy="questionnaires-list-item"]').should('have.length', 1);
 
     cy.contains('Save and continue').click();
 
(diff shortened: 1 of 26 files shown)


cypress/integration/samples.ts
```diff
@@ -93,17 +93,21 @@ context('Samples tests', () => {
...
-    cy.get('[data-cy="questionaries-list-item"]').should('have.length', 1);
+    cy.finishedLoading();
+
+    cy.get('[data-cy="questionnaires-list-item"]').should('have.length', 1);
 
     cy.get('[data-cy="clone"]').click();
+
     cy.contains('OK').click();
-    cy.get('[data-cy="questionaries-list-item"]').should('have.length', 2);
+    cy.get('[data-cy="questionnaires-list-item"]').should('have.length', 2);
```
martin-trajanovski, line 103: Is this the right spelling of questionaries in the plural because when I search online there are so many different spellings 😅 
Screenshot_43

src/components/common/TextFieldWithCounter.tsx
```diff
@@ -27,25 +32,33 @@ const TextFieldWithCounter = (
...
-
-    return classNames.join(' ');
-  };
-
-  const counter = isCounterHidden ? null : (
-    <span className={getCounterClassNames()}>
-      {textLen ? (maxLen ? `${textLen}/${maxLen}` : textLen) : ''}
-    </span>
-  );
+  const counter = isCounterHidden
```
martin-trajanovski, line 35: This is still a bit unreadable when it is hidden with lot of nested conditional operations but I guess it is a bit better than previously

src/components/common/FileUploadComponent.tsx
```diff
@@ -72,7 +72,7 @@ export function FileUploadComponent(props: {
   return (
     <>
       {amountFilesInfo}
-      <List component="ul" className={classes.questionariesList}>
+      <List component="ul" className={classes.questionnairesList}>
```
jekabs-karklins, line 75: Maybe we should stick with "questionaries" for consistency? I'm no language expert but when I looked into it a long time ago both forms seem to be acceptable synonyms "questionnaire" stemming from French, while "questionary" would be a standard English form. Feel free to correct me if I'm wrong :)

jekabs-karklins (approved):

martin-trajanovski (approved):

src/components/common/FileUploadComponent.tsx
peter-asztalos: Well yeah, it is an annoying word, but I wanted to change it to make spell checkers happy (vscode spell checker, Grammarly, Chrome complains about it, probably for historical reasons the 'questionaries' is also acceptable it may just not used in modern texts, just like you use 'your' and not 'thy'), and it also makes it easier to catch misspellings.

For the sake of consistency, we can also go through everything and rename it.

src/components/common/TextFieldWithCounter.tsx
peter-asztalos: I'll add comments in a different PR or see if I can improve the readability a bit.

PR #182, opened 2021-01-13
Swap 1298 remove unneeded questions from review

jekabs-karklins:
## Description

Removing unneeded questions when rendering Questionary overview, and standardizing how questions are rendered in the review.

## Motivation and Context

To better render the question information in the review several components need some special logic. This can now be handled in the QuestionDefinition by specifying the `renderers` property. Additionally, if no renderer is specified then the answer will be omitted from the review completely.

## How Has This Been Tested

- [x] E2E

## Fixes

SWAP-1298

## Changes

QuestionaryComponents and their definitions

## Depends on

N/A


peter-asztalos (approved):
Looks good!

One question: do we need to change anything on the factory side?

jekabs-karklins:
At this point, I don't think we need to change because the logic is already there. but we might need to think about how to prevent the duplication so it's easier to maintain.

PR #202, opened 2021-01-21
UI improvement for details

jekabs-karklins:
## Description

Making QuestionaryDetails consistent and easier to maintain

## Motivation and Context

Questionary often has some additional details to it, like sample information for sample questionary or proposal information for proposal questionary. This commit allows adding the extra details to the questionary view, this making all the presentable information more consistent.

image
image
image

## Changes
QuestionaryDetails

## Depends on
N/A


src/components/review/ProposalQuestionaryReview.tsx
```diff
@@ -1,71 +1,47 @@
...
+    { label: 'Abstract', value: data.abstract },
+    {
+      label: 'Principal Investigator',
+      value: `${data.proposer.firstname} ${data.proposer.lastname}`,
+    },
+    {
+      label: 'Co-Proposers',
+      value: users
+        .map((user: any) => ` ${user.firstname} ${user.lastname}`)
+        .toString(),
```
peter-asztalos, line 35: I think we should use `.join(', ')` here, it would make it more consistent with other parts of the site.

src/components/questionary/QuestionaryDetails.tsx
```diff
@@ -43,10 +62,28 @@ function QuestionaryDetails(
...
+      {title && (
+        <Typography variant="h6" gutterBottom>
+          {title}
+        </Typography>
+      )}
       <Table size="small" {...restProps}>
         <TableBody>
+          {/* Additional details */}
+          {additionalDetails?.map((row, index) =>
+            createTableRow(`additional-detail-${index}`, row)
```
peter-asztalos, line 83: Could we use any part of the `row` for the key instead of the `index`?

src/components/shipments/ShipmentReview.tsx
```diff
@@ -61,48 +52,38 @@ function ShipmentReview({
...
 
+  const additionalDetails: TableRowData[] = [
+    { label: 'Title', value: state.shipment.title },
+    { label: 'Status', value: state.shipment.status },
+    {
+      label: 'Proposal',
+      value: (
+        <Link
+          href="#"
+          onClick={() => history.push(`/ProposalEdit/${proposalData.id}`)}
```
peter-asztalos, line 63: I think we could use the `react-router` `Link` component  so we can open the link like any other (e.g. new tab) and not relying on the `onClick`

peter-asztalos (approved):

PR #206, opened 2021-01-25
SWAP-1169 As a user officer I want to spot dependencies easier

jekabs-karklins:
## Description

This pull-request aims to fix the use-case where the user officer want's to spot question dependencies easier.

## Motivation and Context

The use case is fixed by highlighting the dependee question once it's id is hovered in the depender question.
Also, the dependency can be opened by clicking on it
image

## How Has This Been Tested

- [x] E2E

## Fixes

SWAP-1169

## Depends on

N/A


peter-asztalos (approved):

src/components/template/TemplateQuestionEditor.tsx
```diff
@@ -93,7 +111,7 @@ export default function TemplateQuestionEditor(props: {
       ? '&&'
       : '||';
   const dependencyJsx = dependencies.length ? (
-    <>
+    <div>
```
martin-trajanovski, line 114: Do we need this div element here?

martin-trajanovski (approved):

src/components/template/TemplateQuestionEditor.tsx
jekabs-karklins: Actually, we do, because this div has some styles applied to it.

PR #212, opened 2021-01-26
SWAP-1315 As a user officer I should be able to require numbers to be positive in a number component

peter-asztalos:
## Description

This PR adds support to the number component to restrict values either to only negative or positive numbers.

## How Has This Been Tested

- [x] manual
- [x] e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1315

## Changes

- new constraint fields and validation added
- number component UI fixed (now it's more consistent with other components)
- fixed units field population and now properly registers when the property field value changed during edit

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/142


src/components/common/FormikUICustomSelect.tsx
```diff
@@ -85,7 +85,9 @@ const FormikUICustomSelect = ({
 
   return (
     <>
-      <InputLabel htmlFor={id}>{label}</InputLabel>
+      <InputLabel htmlFor={id} shrink>
```
jekabs-karklins, line 88: Does this mean that the label will be shrunk always (even if the field is empty) , and should we set it everywhere maybe for consistency?

peter-asztalos: Yes, this will shrink the label. Seeing other input elements in the Number field form made me make these changes, so it looks consistent with the rest (the list elements), but I think we will need a similar task in the future to fix the design inconsistencies in the template questionary forms as we did before. We should actually drop the shink and the other elements too, but at least for now, it looks less weird when you have "select" fields. (maybe we should drop the shrink here and fix the other element ¯\_(ツ)_/¯)

src/components/questionary/questionaryComponents/NumberInput/QuestionTemplateRelationNumberInputForm.tsx
```diff
@@ -53,65 +57,93 @@ export const QuestionTemplateRelationNumberForm: FormComponent<QuestionTemplateR
...
-              component={FormikUICustomCheckbox}
-              label="Check to make this field mandatory"
+              name="config.small_label"
+              label="Small label"
+              type="text"
+              component={TextField}
               margin="normal"
               fullWidth
-              InputProps={{ 'data-cy': 'required' }}
+              inputProps={{ 'data-cy': 'small-label' }}
```
martin-trajanovski, line 78: I see somewhere we use `InputProps` and here it starts with lowercase `inputProps`. Is it working in both cases and can we use the same form almost everywhere?

martin-trajanovski (approved):

src/components/questionary/questionaryComponents/NumberInput/QuestionTemplateRelationNumberInputForm.tsx
peter-asztalos: There is some difference, according to the documentation:
```
inputProps | object |   | Attributes applied to the input element.
InputProps | object |   | Props applied to the Input element. It will be a FilledInput, OutlinedInput or Input component depending on the variant prop value.
```

Just we have some custom components so it's inconsistent as some properties are either not passed or missing the right types from the props.

Looking at the documentation we should use the lower case version as we want to set a custom attribute on the element, the upper case version can be used to set React props.

src/components/common/FormikUICustomSelect.tsx
peter-asztalos: I double-checked it, looks like without the shrink it just uses the normal font size but stays in the upper corner instead of coming down and filling out the input field area like other components do. This change makes it looks less weird.

Without shrink: 
Screenshot 2021-02-01 114718

WIth shrink:
Screenshot 2021-02-01 114833

Scott-James-Hurley (approved):

PR #216, opened 2021-02-01
SWAP-1330 As a user I should have better indication of time when a PDF is downloading

peter-asztalos:
## Description

This PR changes how the PDF/XLSX reports are downloaded by keeping track of download requests and showing a list to the user of them.

## Motivation and Context

The users haven't had any kind of feedback on the UI when trying to download PDF/XLSX.

## How Has This Been Tested

- [x] e2e test
- [x] manual test

## Fixes

https://jira.esss.lu.se/browse/SWAP-1330

## Changes

- show a small box to the user with the list of files they requested to download
- `graphql-request` upgraded

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/146

### DIFF
--- a/src/hooks/SEP/useDownloadXLSXSEP.ts
+++ b/src/hooks/SEP/useDownloadXLSXSEP.ts
@@ -1,18 +1,17 @@
-import { useCallback } from 'react';
+import { useCallback, useContext } from 'react';
 
+import {
+  DownloadContext,
+  PREPARE_DOWNLOAD_TYPE,
+} from 'context/DownloadContextProvider';
 export function useDownloadXLSXSEP() {
-  const downloadSEPXLSX = useCallback((sepId, callId) => {
-    const element = document.createElement('a');
-    element.setAttribute('href', `/download/xlsx/sep/${sepId}/call/${callId}`);
-    element.setAttribute('download', 'download');
-
-    element.style.display = 'none';
-    document.body.appendChild(element);
-
-    element.click();
-
-    document.body.removeChild(element);
-  }, []);
+  const { prepareDownload } = useContext(DownloadContext);
+  const downloadSEPXLSX = useCallback(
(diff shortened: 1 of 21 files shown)


jekabs-karklins (approved):

Scott-James-Hurley (approved):

src/hooks/SEP/useDownloadXLSXSEP.ts
```diff
@@ -1,18 +1,17 @@
...
-    document.body.appendChild(element);
-
-    element.click();
-
-    document.body.removeChild(element);
-  }, []);
+  const { prepareDownload } = useContext(DownloadContext);
+  const downloadSEPXLSX = useCallback(
+    (sepId: number, callId: number, name: string) => {
+      prepareDownload(PREPARE_DOWNLOAD_TYPE.XLSX_SEP, [[sepId, callId]], name);
```
martin-trajanovski, line 11: I am not really sure why we need here `Array<number[]>` for sending `sepId` and `callId`. As I see the acceptable type is `Array<number | number[]>` and here it is a bit confusing that we send `[[sepId, callId]]` and we can send just `[sepId, callId]`. Otherwise it looks great!

martin-trajanovski (approved):

src/hooks/SEP/useDownloadXLSXSEP.ts
peter-asztalos: `sepId` and `callId` together identify the resource we want to download, separating them would mean we want to download 2 files, not one. In most cases, we send the array of IDs, as we want to download multiple resources, but with SEP you need two different IDs to identify one, so we group them.
`useDownloadXLSXSEP` hides this part and the `generateLink` extracts the IDs for us.

martin-trajanovski: Great! Good to know.

PR #224, opened 2021-02-03
Swap 1283 api access

martin-trajanovski:
## Description

Functionality that allows user officers to create/update/delete API access tokens with specific permissions for external services to be able to read/modify data.

## Motivation and Context

This change is because of a need for multiple external services to mainly read but also modify data in user office.

## How Has This Been Tested

- e2e tests
- Manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-1283
https://jira.esss.lu.se/browse/SWAP-1338

## Changes

Created new route and all its components to be able to list/create/update/delete access tokens.

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/156


src/components/common/InputDialog.tsx
```diff
@@ -14,7 +14,7 @@ function InputDialog(props: DialogProps) {
   const classes = useStyles();
 
   return (
-    <Dialog maxWidth="sm" {...props}>
+    <Dialog maxWidth={props.maxWidth || 'sm'} {...props}>
```
jekabs-karklins, line 17: I wonder if we have to have this change, because if value maxWidth is in props then it will be overridden as comes after the attribute. No?

But if we have to have it then I think it is better to write {props.maxWidth ?? 'sm'} (with nullish coalescing)

src/components/questionary/QuestionaryDetails.tsx
```diff
@@ -30,13 +30,15 @@ export interface TableRowData {
 }
 function QuestionaryDetails(
   props: {
-    questionaryId: number;
+    questionaryId: number | undefined;
```
jekabs-karklins, line 33: I think that `undefined` is not a valid questionaryId value here. Do we need to have it like this?

src/components/settings/apiAccessTokens/CreateUpdateApiAccessToken.tsx
```diff
@@ -0,0 +1,343 @@
...
+    maxHeight: '550px',
+    overflowY: 'auto',
+    overflowX: 'hidden',
+  },
+  submit: {
+    margin: theme.spacing(3, 0, 2),
+  },
+  error: {
+    color: theme.palette.error.main,
+    marginRight: '10px',
```
peter-asztalos, line 45: You using `theme.spacing(...)` in other places, I think this and everything related to spacing should use it (margin*, padding*).

src/components/settings/apiAccessTokens/CreateUpdateApiAccessToken.tsx
```diff
@@ -0,0 +1,343 @@
...
+
+  const normalizeAccessPermissions = (data: string | undefined) => {
+    const permissionsArray: string[] = [];
+
+    if (data) {
+      const parsedPermissions = JSON.parse(data);
+
+      for (const key in parsedPermissions) {
+        if (Object.prototype.hasOwnProperty.call(parsedPermissions, key)) {
+          permissionsArray.push(key);
```
peter-asztalos, line 113: This `for` and `if` can be simplified to `permissionsArray = Object.keys(parsedPermissions)`

src/components/settings/apiAccessTokens/CreateUpdateApiAccessToken.tsx
```diff
@@ -0,0 +1,343 @@
...
+      }
+    }
+
+    return permissionsArray;
+  };
+
+  const initialValues: FormPermissionsWithAccessToken = apiAccessToken
+    ? {
+        ...apiAccessToken,
+        accessToken: `Bearer ${apiAccessToken.accessToken}`,
```
peter-asztalos, line 124: Do we need the bearer part here?

src/hooks/admin/useApiAccessTokensData.ts
```diff
@@ -0,0 +1,47 @@
...
+    api()
+      .getAllApiAccessTokensAndPermissions()
+      .then(data => {
+        if (data.allAccessTokensAndPermissions) {
+          setApiAccessTokens(
+            data.allAccessTokensAndPermissions as PermissionsWithAccessToken[]
+          );
+        }
+        setLoadingApiAccessTokens(false);
+      });
```
peter-asztalos, line 39: Please here and in the other hook check if the component was unmounted before updating the state.

src/components/settings/apiAccessTokens/CreateUpdateApiAccessToken.tsx
martin-trajanovski: I think here yes because I would like to show it inside the input to be able just to copy directly as a whole useful token. I have that copy button for the input so you can just copy the token all together and share it with someone who needs to use it.

src/components/questionary/QuestionaryDetails.tsx
martin-trajanovski: Let me double check this I think I did it mainly because typescript was complaining about some modifications in the backend after I changed some resolvers to be nullable: true

martin-trajanovski: I made the questionary nullable in the resolvers because it was needed for example if you do not have access to get questionary on a proposal but just proposal data it needs to be nullable.

peter-asztalos (approved):

PR #226, opened 2021-02-04
SWAP-1344 & SWAP-1342: Missing call, instrument filter for instrument scientist

peter-asztalos:
## Description

 This PR fixes the instrument filter for instrument scientists and adds the missing calls filter.

## How Has This Been Tested

- [x] e2e tests
- [x] manual test

## Fixes

https://jira.esss.lu.se/browse/SWAP-1342
https://jira.esss.lu.se/browse/SWAP-1344

## Changes

- instrument filter fixed
- missing calls filter added

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/159

### DIFF
--- a/cypress/integration/instruments.ts
+++ b/cypress/integration/instruments.ts
@@ -1,9 +1,167 @@
 context('Instrument tests', () => {
   const faker = require('faker');
 
+  function createInstrument({
+    name,
+    shortCode,
+    description,
+  }: {
+    name: string;
+    shortCode: string;
+    description: string;
+  }) {
+    cy.contains('Create').click();
+    cy.get('#name').type(name);
+    cy.get('#shortCode').type(shortCode);
+    cy.get('#description').type(description);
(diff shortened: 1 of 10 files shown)


cypress/integration/instruments.ts
```diff
@@ -1,9 +1,167 @@
 context('Instrument tests', () => {
   const faker = require('faker');
 
+  function createInstrument({
```
martin-trajanovski, line 4: Isn't it better to add these functions as commands or they are used just in this file? I guess we use createInstrument and createCall in other files as well.

martin-trajanovski (approved):

cypress/integration/instruments.ts
peter-asztalos: In long term that would be the goal, but for now I didn't bother with it. Maybe in a dedicated task.

jekabs-karklins (approved):

PR #227, opened 2021-02-04
SWAP-1251: Add table for units and remove old static unit file

bolmsten:
## Description

A user officer can now create units int the GUI to use in number and interval components.

## Motivation and Context

Allow user officer to change units without need for development effort

## How Has This Been Tested

E2E test

## Fixes

SWAP-1251

## Changes

Changes interval and number component to use units fetched from backend. It also adds a new table for adding and deleting units

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/160


src/components/MenuItems.tsx
```diff
@@ -240,6 +240,12 @@ const SettingsMenuListItem = () => {
         <ListItemText primary="Settings" />
       </ListItem>
       <Collapse in={isExpanded} timeout="auto" unmountOnExit>
+        <ListItem component={NavLink} to="/Units" button>
+          <ListItemIcon>
+            <ProposalSettingsIcon />
```
martin-trajanovski, line 245: Are we using the same icon for both Proposal statuses and Units?

src/components/questionary/questionaryComponents/Interval/QuestionNumberInputForm.tsx
```diff
@@ -0,0 +1,117 @@
+import makeStyles from '@material-ui/core/styles/makeStyles';
```
martin-trajanovski, line 1: Not sure why this `QuestionNumberInputForm` is inside `Interval` folder. I see the same file name inside `NumberInput` form. Or maybe it is not the same but correct me if I am wrong.

src/components/settings/unitList/UnitTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+        data={units}
+        isLoading={loadingUnits}
+        options={{
+          search: true,
+          debounceInterval: 400,
+        }}
+        urlQueryParams={urlQueryParams}
+        setUrlQueryParams={setUrlQueryParams}
+        actions={[
+          rowActionData => {
```
martin-trajanovski, line 80: I think we shouldn't use custom delete functionality. You just need to use existing `SuperMaterialTable` delete property. Something like `delete={deleteUnit}` and define the `deleteUnit` function.

martin-trajanovski (changes requested):

jekabs-karklins (approved):
Looks good.
One thing I see is that you removed physical property and only left the unit, which is simpler, and also allows to specify units of multiple properties in the same question. but that also raises the question of how will we do a search on that question type. 

Say I want to find all samples where sample weight is greater than 1 kg.

For these types of query, we will need to
1, Add the unit back. We can just have another column in the units table called "property" (easy)
2. For every property, we need a conversion formula to SI units, (medium)
3. Figure out how to tackle the query where multiple physical property units are allowed. (hard)

src/components/MenuItems.tsx
bolmsten: Fixed,

src/components/settings/unitList/UnitTable.tsx
bolmsten: Done

src/components/questionary/questionaryComponents/Interval/QuestionNumberInputForm.tsx
bolmsten: removed unused file

martin-trajanovski (approved):

RasmiaKulan (approved):

PR #230, opened 2021-02-08
SWAP-1350 Add boxes for selecting reviewers

peter-asztalos:
## Description

This PR updates the "member assignment to a proposal" modal, letting us select and assign multiple members in one go.

## Motivation and Context

Right now you can only select one user per modal making the process a bit dull.

## How Has This Been Tested

- [x] e2e tests
- [x] manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-1350

## Changes

- replace the modal component
- update the assignment part

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/168


cypress/integration/SEPs.ts
```diff
@@ -546,12 +546,14 @@ context('Scientific evaluation panel tests', () => {
     cy.get('[role="dialog"]')
       .contains(sepMembers.reviewer.surname)
       .parent()
-      .find('[title="Add reviewer"]')
+      .find('input[type="checkbox"]')
```
jekabs-karklins, line 549: Why can't we keep using the title? I think using a title is easier to read and maintain

jekabs-karklins (approved):

cypress/integration/SEPs.ts
peter-asztalos: I think the checkbox there has no "title", I can try to add a generic way where you can set it, it just the `PeopleTable` is used at many places already with this solution.

martin-trajanovski (approved):

PR #233, opened 2021-02-11
Swap 1346 7 technical review small fixes

martin-trajanovski:
## Description

Add ability to submit technical review and prevent updates if submitted. Also prevent leaving the form if it is not saved.

## Motivation and Context

Previously there was no possibility to submit a technical review only update and we need a way to submit and lock it.

## How Has This Been Tested

e2e tests on the frontend

## Fixes

https://jira.esss.lu.se/browse/SWAP-1346
https://jira.esss.lu.se/browse/SWAP-1347

## Changes

Adding submit button and it's functionality. Showing confirmation if you try to leave technical review form with unsaved changes.


src/components/review/ProposalGrade.tsx
```diff
@@ -126,7 +126,7 @@ export default function ProposalGrade(props: {
               color="secondary"
               onClick={() => {
                 const confirmed = window.confirm(
-                  'I am aware that no futhure changes to the grade is possible after submission.'
+                  'I am aware that no future changes to the grade is possible after submission.'
```
jekabs-karklins, line 129: I think this is actually supposed to be "further"

src/components/review/ProposalTechnicalReview.tsx
```diff
@@ -1,64 +1,121 @@
...
-          props.setReview({
-            proposalID: props?.data?.proposalID,
-            timeAllocation: +values.timeAllocation,
-            comment: values.comment,
-            publicComment: values.publicComment,
-            status:
-              TechnicalReviewStatus[values.status as TechnicalReviewStatus],
-          } as CoreTechnicalReviewFragment);
+          if (shouldSubmit) {
+            const confirmed = window.confirm(
```
jekabs-karklins, line 101: I think window.confirm() is sufficient here, but just wanted to note that there is also a HOC 
 withConfirm https://github.com/UserOfficeProject/user-office-frontend/blob/develop/src/utils/withConfirm.tsx

jekabs-karklins (approved):

peter-asztalos (approved):

src/components/review/ProposalTechnicalReview.tsx
martin-trajanovski: I was thinking about this but I know that this technical review thing might change because we mentioned that we like the full screen modal sliding from bottom to top like we have for SEP reviews. If this is the case I know that you mentioned once it is a bit hard to make this confirm on top of another modal but if I am wrong I can try to change it and make it work.

martin-trajanovski: I fixed it for now and it uses `withConfirm` HOC I think it is better like this at the moment thanks.

PR #241, opened 2021-02-17
Swap 1348 update status proposal table

martin-trajanovski:
## Description

Returning the next proposal status if exists on `assignProposalToSEP` mutation, so we can update the frontend proposals table immediately if it is successful. 

## Motivation and Context

The proposal officer table in the frontend wasn't updated at real time when there was proposal workflow status change.

## How Has This Been Tested

- manual testing
- e2e tests on the frontend

## Fixes

https://jira.esss.lu.se/browse/SWAP-1348

## Changes

Update the proposal officer table with the new proposal status right after proposal is assigned to a SEP without a reload.


peter-asztalos (approved):
Looks good (☞ﾟヮﾟ)☞

PR #244, opened 2021-02-22
SWAP-1395: Add delete button for calls

bolmsten:
## Description

Add button for deletion of a call

## Motivation and Context

A user officer should be able to delete a call that has no dependencies 

## How Has This Been Tested

E2E

## Fixes

SWAP-1395

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/181

- [X] I have added tests to cover my changes.


cypress/integration/calls.ts
```diff
@@ -308,7 +308,7 @@ context('Calls tests', () => {
       .click();
 
     cy.get('[title="Delete"]')
-      .first()
+      .eq(1)
```
martin-trajanovski, line 311: Maybe this can be targeted better and instead of getting second delete button in the view we can have something like: 
`cy.get('some_better_selector [title="Delete"]').first().click()`

cypress/integration/instruments.ts
```diff
@@ -644,7 +644,7 @@ context('Instrument tests', () => {
       .click();
 
     cy.get('[title="Delete"]')
-      .first()
+      .eq(1)
```
martin-trajanovski, line 647: Maybe this can be targeted better and instead of getting second delete button in the view we can have something like: 
`cy.get('some_better_selector [title="Delete"]').first().click()`

martin-trajanovski (changes requested):
Otherwise looks good!

cypress/integration/instruments.ts
martin-trajanovski: I see this one can be used for example: `cy.get('[data-cy="call-instrument-assignments-table"] [title="Delete"]').first().click()`

cypress/integration/calls.ts
bolmsten: I agree that is much better

martin-trajanovski (approved):

PR #259, opened 2021-03-01
fix icons in multiple rows and some settings test improvements

martin-trajanovski:
## Description

Small styling fix for more than 3 actions on user officer proposals table and potential fix for settings randomly failing tests.

### DIFF
--- a/cypress/integration/settings.ts
+++ b/cypress/integration/settings.ts
@@ -75,9 +75,7 @@ context('Settings tests', () => {
 
       lastPageButtonElement.click({ force: true });
 
-      cy.get('[title="Edit"]')
-        .last()
-        .click();
+      cy.get('[title="Edit"]').last().click();
 
       cy.get('#shortCode').should('be.disabled');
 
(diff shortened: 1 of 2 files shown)


cypress/integration/settings.ts
```diff
@@ -410,7 +392,7 @@ context('Settings tests', () => {
 
       cy.get('.MuiTable-root tbody tr')
         .first()
-        .then(element => expect(element.text()).to.contain('submitted'));
+        .then((element) => expect(element.text()).to.contain('submitted'));
```
jekabs-karklins, line 395: Why do we need parenthesis here? Are they added by linting auto fix?

jekabs-karklins (approved):
Remember to not merge today

cypress/integration/settings.ts
martin-trajanovski: Yes this is a change in the prettier that was introduced after we updated it. https://prettier.io/docs/en/options.html#arrow-function-parentheses

PR #260, opened 2021-03-02
fix initial value for rich text input

bolmsten:
## Description

Before the input that a user added to a rich text question did not render when revisited as it had been wrongly initialised. 

## Motivation and Context

Fix bug


jekabs-karklins (approved):
Looks good

PR #261, opened 2021-03-02
Swap 1406 technical review submit

martin-trajanovski:
## Description

User officer should be able to re-open submitted technical review.

## Motivation and Context

Previously when technical review was submitted there was no way to reopen it.>

## How Has This Been Tested

manual and e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1406

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/197


cypress/integration/SEPs.ts
```diff
@@ -12,7 +12,7 @@ function readWriteReview() {
 
   cy.finishedLoading();
 
-  cy.get('@dialog').contains('Proposal information');
+  cy.get('@dialog').contains('Proposal Information');
```
jekabs-karklins, line 15: You could do `cy.get('@dialog').contains('Proposal information', { matchCase: false });` so it's easier to maintain, unless you want to test the case sensitivity

src/components/common/TabPanel.tsx
```diff
@@ -40,55 +41,53 @@ function a11yProps(index: number) {
...
+  });
 
   const handleChange = (
     event: React.ChangeEvent<Record<string, unknown>>,
     newValue: number
   ) => {
-    if (props.shouldPreventTabChange && value !== newValue) {
-      if (
-        window.confirm(
-          'Changes you recently made in this tab will be lost! Are you sure?'
```
jekabs-karklins, line 59: Is this removed because <PromptIfDirty /> is already taking care of confirmation?

jekabs-karklins (approved):

cypress/integration/SEPs.ts
martin-trajanovski: Yes maybe that is even better option. I just found that test failing randomly and I think it was because of not matching the case.

src/components/common/TabPanel.tsx
martin-trajanovski: This is now handled by `<Prompt />` component since we are changing the url if you try to leave without saving the form. It is cleaner and nicer way of doing things. Previously I did a custom thing because when you change tabs nothing was changed in the url and I was checking manually if there is a change to prevent changing the tab.

PR #262, opened 2021-03-02
Swap 1426 -- Allow SEP Chair and Secretary to see and download proposal

bolmsten:
## Description

A SEP Chair or Secretary needs to see proposals before assignment to reviewers. This PR adds both a view in the SEP "Proposal and assignments" tab and a download button


src/components/SEP/MeetingComponents/ProposalViewModal/SEPProposalViewModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+const Transition = React.forwardRef<unknown, TransitionProps>(SlideComponent);
+
+type SEPProposalViewModalProps = {
+  proposalViewModalOpen: boolean;
+  proposalId: number;
+  setProposalViewModalOpen: (isOpen: boolean) => void;
+};
+
+const SEPProposalViewModal: React.FC<SEPProposalViewModalProps> = ({
```
martin-trajanovski, line 47: I guess it is fine for now but this is almost similar to the `ProposalReviewModal` and maybe later we can refactor a bit and have one more reusable modal where we can see proposal information that we want by passing different tabs as parameters.

martin-trajanovski (approved):

PR #264, opened 2021-03-03
Refactoring NavigationFragment

jekabs-karklins:
## Description

This pull request aims to make NavigationFragment more reusable. 

## Motivation and Context

NavigationFragment was too specific to the proposal questionary. Since NavigationFragment is used by different types of questionnaires it should be possible to specify other kinds of buttons in the NavigationFragment without hacking it. 

## How Has This Been Tested

- [x] E2E

## Fixes

N/A

## Changes

NavigationFragment, and places using it

## Depends on

N/A


src/components/questionary/Questionary.tsx
```diff
@@ -34,6 +34,7 @@ const useStyles = makeStyles((theme) => ({
   },
   root: {
     width: 'inherit',
+    minWidth: '500px',
```
peter-asztalos, line 37: This one once added and removed in (https://github.com/UserOfficeProject/user-office-frontend/blob/80d00917969c87bfaa3b2d3bf4f8971fae43209d/src/components/questionary/Questionary.tsx#L36), now it's back, I think we should talk about this part or leave a comment why is it needed.

peter-asztalos (changes requested):
Looks good but I think the formatting is wrong: after upgrading prettier parentheses are required for arrow function params. A suggestion: do a npm ci and then run npm lint and lint:fix to see the formatting issues.

src/components/questionary/Questionary.tsx
jekabs-karklins: I think you mean L:37 minWidth?
That is if questions are short the view looks squashed. It is the case for shipment review where there are just few questions, but could be the case for other questionnaires as well.

peter-asztalos: I meant why is it needed there, so others won't remove it in the future like they did previously.

peter-asztalos (approved):

PR #266, opened 2021-03-04
SWAP-1393 Order disappears on proposal table after row selection

peter-asztalos:
## Description

This PR fixes the ordering reset that happens when the user selects a row.

## How Has This Been Tested

- [x] manual test

## Fixes

- https://jira.esss.lu.se/browse/SWAP-1393

## Changes

- order reset fixed


package.json
```diff
@@ -50,6 +50,7 @@
     "react-beautiful-dnd": "^13.0.0",
     "react-cookie": "^4.0.3",
     "react-dom": "^17.0.0",
+    "react-fast-compare": "^3.2.0",
```
martin-trajanovski, line 53: I didn't know about this package but it is good that you introduced it so we can use it in the future where we need.

martin-trajanovski (approved):

jekabs-karklins (approved):

PR #272, opened 2021-03-08
SWAP-1400: Add deletion of SEPs and filtering

bolmsten:
## Description

Add buttons for deletion of SEP and filter for active and inactive

## Motivation and Context

user office should only see active SEP's as default and have the possibility to delete empty SEP's

## How Has This Been Tested

E2E tests

## Fixes

SWAP-1400

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/207


Scott-James-Hurley (approved):

src/components/SEP/SEPStatusFilter.tsx
```diff
@@ -0,0 +1,55 @@
+import FormControl from '@material-ui/core/FormControl';
```
martin-trajanovski, line 1: As far as I know we have this similar active/inactive filter for calls but not sure if that can be reused or not.

martin-trajanovski (approved):

PR #274, opened 2021-03-09
SWAP-1306 As a user I want to see when my proposal has been scheduled

peter-asztalos:
## Description

This PR adds support for Users to see their upcoming (or all) beam times.

## How Has This Been Tested

- [x] manual
- [x] e2e test 

## Fixes

https://jira.esss.lu.se/browse/SWAP-1306

## Changes

- support to access scheduler
- core and scheduler github actions updated
- core and scheduler docker compose files updated
- API gateway updated


cypress/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM cypress/included:6.5.0
+FROM cypress/included:6.6.0
```
martin-trajanovski, line 1: I see there is already version 6.8.0 but maybe we could update that next time to latest.

cypress.json
```diff
@@ -1,4 +1,4 @@
 {
   "video": false,
-  "baseUrl": "http://127.0.0.1"
+  "baseUrl": "http://127.0.0.1:3000"
```
martin-trajanovski, line 3: I guess there wont be any problem to merge this. If so we do not need to change it locally.

package.json
```diff
@@ -59,7 +59,7 @@
     "use-query-params": "^1.1.9",
     "yup": "~0.29.3"
   },
-  "proxy": "http://localhost:4000",
+  "proxy": "http://localhost:4100",
```
martin-trajanovski, line 62: Not sure why this change is needed from 4000 to 4100

martin-trajanovski (approved):
Otherwise it looks great!

package.json
peter-asztalos: 4100 is the address of the api gateway

cypress/Dockerfile
peter-asztalos: Yes, unless there is any major issue with 6.6.0 we can upgrade next time.

jekabs-karklins (approved):

PR #284, opened 2021-03-15
#21: Update logic behind the proposal being read only

Ziggiyzoo:
## Description
Closes https://github.com/UserOfficeProject/stfc-user-office-project/issues/80 

Sorry about these links I don't seem to be able to use the standard `#80` as it keeps taking me to another issue?

Updates logic behind using workflows to allow users to edit proposals after their submission.

In order for this to work:

-> User Officer creates a Proposal workflow where the status is changed to `EDITABLE_SUBMISSION` from `DRAFT` after the proposal is submitted.
-> User Officer creates a Call that uses this proposal workflow.
-> User must create their proposal using the created call.

In the case of STFC and ESS, STFC would remove the default call so the user can only use the created one while ESS would keep the call as default.

## Motivation and Context

STFC wishes to allow its users to edit their proposals after submission, but before the round has ended. And ESS wishes to keep the original workflow.

## Changes

Adds in logic to determine the status of the Proposal. As well as changes the way the `isReadonly` status is determined. 

## Depends on

Depends on https://github.com/UserOfficeProject/user-office-backend/pull/217 as this is the backed change that adds the new `EDITABLE_SUBMISSION` status to the StatusesTable DB patch.


src/components/proposal/ProposalContainer.tsx
```diff
@@ -71,9 +71,23 @@ const proposalReducer = (
...
 const isProposalSubmitted = (proposal: { submitted: boolean }) =>
   proposal.submitted;
 
+//Is the proposal in draft or edited state? If so set read only to false.
+function isProposalEditable(proposal: { status: ProposalStatus }) {
+  if (
+    proposal.status.shortCode.toString() === 'DRAFT' ||
+    proposal.status.shortCode.toString() === 'EDITABLE_SUBMISSION'
+  ) {
+    return false;
```
jekabs-karklins, line 85: I think this should be true

src/components/proposal/ProposalContainer.tsx
```diff
@@ -71,9 +71,23 @@ const proposalReducer = (
...
+  } else {
+    return true;
```
jekabs-karklins, line 87: This false

src/components/proposal/ProposalContainer.tsx
```diff
@@ -104,7 +116,7 @@ const createReviewWizardStep = (): WizardStep => ({
       title: 'Review',
       isCompleted: isProposalSubmitted(proposalState.proposal),
       isReadonly:
-        isProposalSubmitted(proposalState.proposal) ||
+        isProposalEditable(proposalState.proposal) &&
```
jekabs-karklins, line 119: And this negated

jekabs-karklins (commented):
Looks good

src/components/proposal/ProposalContainer.tsx
```diff
@@ -76,9 +76,24 @@ const proposalReducer = (
   return draftState;
 };
 
+//Option 1: Change isProposalSubmitted to isProposalEditable and have this be true when proposal status is either DRAFT or EDITABLE_SUBMISSION
+//Option 2: Have the isProposalSubmitted bypassed if the proposal status is EDITABLE_SUBMISSION
 const isProposalSubmitted = (proposal: { submitted: boolean }) =>
   proposal.submitted;
 
+//Is the proposal in draft or edited state? If so set read only to false.
+function isProposalEditable(proposal: ProposalSubsetSubmission) {
```
mr-field, line 85: I also think the return values should be swapped, or alternatively this function should be renamed to `isReadOnly`.

davidsymons1999: I think the renaming to `isReadOnly` would make more sense, but I don't know how that will react with the fact that it would be called like:
`isReadOnly: isReadOnly(proposal)`

Whereas incersing the True/False would require `isProposalEditable()` to become `!isProposalEditable()` in some cases?

mr-field: I think having something like `isReadOnly: isReadOnly(proposal)` should be fine, as the first `isReadOnly` is the property of an object while the second is a function, so they're in different scopes.

mr-field (approved):

jekabs-karklins (approved):

PR #285, opened 2021-03-16
Refactor template forms

jekabs-karklins:

### DIFF
--- a/cypress/integration/templates.ts
+++ b/cypress/integration/templates.ts
@@ -78,9 +78,7 @@ context('Template tests', () => {
 
     cy.contains('Add Text Input').click();
 
-    cy.get('[data-cy=question]')
-      .clear()
-      .type(textQuestion);
+    cy.get('[data-cy=question]').clear().type(textQuestion);
 
     cy.contains('Is required').click();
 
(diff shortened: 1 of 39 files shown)


cypress/integration/templates.ts
```diff
@@ -537,6 +537,8 @@ context('Template tests', () => {
 
     cy.contains('Save').click();
 
+    cy.wait(1000);
```
martin-trajanovski, line 540: Can't you just use `finishedLoading` or `notification` here instead of wait?

src/components/questionary/QuestionaryComponentRegistry.tsx
```diff
@@ -26,17 +26,27 @@ import { sampleDeclarationDefinition } from './questionaryComponents/SampleDecla
...
 
-export interface FormProps<ValueObjectType> {
-  field: ValueObjectType;
+export type FormChildren<ValueObjectType> = (
+  formikProps: FormikProps<ValueObjectType>
+) => React.ReactNode;
+
+export interface QuestionFormProps {
+  question: Question;
+  closeMe: () => unknown;
```
martin-trajanovski, line 35: Is the return type here on all functions `unknown` for sure or it is `void`?

martin-trajanovski (approved):
Looks good! 👀 👍

src/components/questionary/QuestionaryComponentRegistry.tsx
jekabs-karklins: Unknown because it's a callback function and does not care about the return type

PR #288, opened 2021-03-18
Swap 768 As a user officer I want to moderate questions so that I can keep my templates and question repository clean

jekabs-karklins:
## Description

This PR is a backend part for being able to show all questions in one place for moderation purposes.

image

## Motivation and Context

This page shows how many proposals/samples/shipments has been created using the question and also in how many templates it has been used.

## Fixes

SWAP-768

## Changes

Added new page

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/220


cypress/integration/questions.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  it('User officer search questions', () => {
+    cy.login('officer');
+
+    cy.navigateToTemplatesSubmenu('Proposal templates');
+
+    cy.contains('default template')
+      .parent()
+      .get("[title='Edit']")
+      .first()
```
martin-trajanovski, line 23: I don't think we need `.first()` here because if I am correct you are getting the template by name and then parent element should have only one edit.

cypress/integration/questions.ts
```diff
@@ -0,0 +1,65 @@
...
+      .click();
+
+    cy.get('[data-cy=show-more-button]').last().click();
```
martin-trajanovski, line 26: Maybe we should avoid using `.last()` or `.first()` as much as possible. It is easier to maintain it later if it is better targeted.

martin-trajanovski (approved):

src/hooks/template/useQuestions.ts
```diff
@@ -0,0 +1,33 @@
...
+  const [questions, setQuestions] = useState<QuestionWithUsage[]>([]);
+  const [loadingQuestions, setLoadingQuestions] = useState(true);
+
+  const api = useDataApi();
+
+  useEffect(() => {
+    setLoadingQuestions(true);
+    api()
+      .getQuestions({ filter: questionsFilter })
+      .then((data) => {
```
peter-asztalos, line 19: Please add a check to see if the component is unmounted or not.

peter-asztalos (changes requested):

src/hooks/template/useQuestions.ts
martin-trajanovski: Yes I think we should do that to all hooks that are using useEffect to prevent those console warnings.

peter-asztalos (approved):

src/generated/sdk.ts
```diff
@@ -318,7 +318,7 @@ export type Entry = {
 export type Equipment = {
   __typename?: 'Equipment';
   id: Scalars['ID'];
-  owner: Maybe<User>;
+  owner: User;
```
peter-asztalos, line 321: I think your local scheduler images are out-of-date and that's why the pipeline fails.

PR #291, opened 2021-03-22
SWAP-1449 Fix for rendering interval answer #resolved

jekabs-karklins:
## Description

Fix for rendering interval answer

## Motivation and Context

Previously when min was 0 or left blank, the answer was rendered as "Left blank"


peter-asztalos (approved):
Do we have to make changes on the factory side?

martin-trajanovski (approved):

jekabs-karklins:
> Do we have to make changes on the factory side?

No, this is more of a formatting fix

PR #293, opened 2021-03-22
SWAP-1496 The beamtimes do not show up for the PI. They should show up and they should also show up for the co-proposer

peter-asztalos:
## Description

This PR includes the e2e test for SWAP-1496 

## How Has This Been Tested

- [x] e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1496

## Changes

- e2e test updated

## Depends on

https://github.com/UserOfficeProject/user-office-scheduler-backend/pull/23

### DIFF
--- a/cypress/integration/scheduler.ts
+++ b/cypress/integration/scheduler.ts
@@ -36,45 +36,56 @@ context('Scheduler tests', () => {
     cy.finishedLoading();
 
     cy.contains('Upcoming beam times').should('not.exist');
+    cy.logout();
   });
 
-  it('User should be able to see upcoming beam times', () => {
-    {
-      const query = `
+  it('User should not be able to see upcoming beam times in DRAFT state', () => {
+    const query = `
       mutation bulkUpsertScheduledEvents($input: BulkUpsertScheduledEventsInput!) {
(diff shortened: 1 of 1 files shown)


cypress/integration/scheduler.ts
```diff
@@ -36,61 +36,101 @@ context('Scheduler tests', () => {
     cy.finishedLoading();
 
     cy.contains('Upcoming beam times').should('not.exist');
+    cy.logout();
```
peter-asztalos, line 39: For some weird reasons the user stays logged in, couldn't figure out why, so I added an explicit logout

jekabs-karklins (approved):

cypress/integration/scheduler.ts
martin-trajanovski: Maybe it has to do something with the cypress upgrade or it is not because of that what do you think?

martin-trajanovski (approved):

cypress/integration/scheduler.ts
peter-asztalos: I think it's related to the request we make to change the booking status, either a bug or something else.

PR #294, opened 2021-03-23
Swap 1485 instrument scientist can not search questions

jekabs-karklins:
## Description

This PR aims to fix a bug where IS can not filter proposals by question answers 

## Motivation and Context

IS should be able to filter proposals by question answers

## Fixes

SWAP-1485

## Depends on

N/A

### DIFF
--- a/cypress/integration/instruments.ts
+++ b/cypress/integration/instruments.ts
@@ -1,7 +1,6 @@
 context('Instrument tests', () => {
   const faker = require('faker');
-  const questionText = faker.lorem.words(3)
-
+  const questionText = faker.lorem.words(3);
 
   function createInstrument({
     name,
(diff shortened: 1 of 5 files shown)


cypress/integration/instruments.ts
```diff
@@ -161,15 +163,34 @@ context('Instrument tests', () => {
 
     cy.createCall(call2);
 
+    cy.navigateToTemplatesSubmenu('Proposal templates');
+
+    cy.contains('default template')
+      .parent()
+      .get("[title='Edit']")
+      .first()
```
martin-trajanovski, line 171: I think you don't need `.first()` here at all. There is only one edit button when you get it by parent.

cypress/integration/instruments.ts
```diff
@@ -161,15 +163,34 @@ context('Instrument tests', () => {
...
+      .click();
+
+    cy.createTopic('Topic for questions');
+
+    cy.get('[data-cy=show-more-button]').last().click();
```
martin-trajanovski, line 176: It would be much better if we can target this item much better instead of using `.last()` and the same goes for the one below

martin-trajanovski (approved):
Otherwise it looks good! 👀 👍

peter-asztalos (approved):

PR #301, opened 2021-03-30
resolve some console errors when changing roles

martin-trajanovski:
## Description

Preventing re-rendering inside some hooks and fixed some `console` that were fired when you try to change user role.


src/components/AppToolbar/RoleSelection.tsx
```diff
@@ -35,7 +35,12 @@ const RoleSelection: React.FC<RoleSelectionProps> = ({ close }) => {
       }
     };
     getUserInformation();
-  }, [user.id, api]);
+
+    return () => {
+      unmounted = true;
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
jekabs-karklins, line 43: I wonder does passing in [user.id, api] cause any issues?

jekabs-karklins (approved):

src/components/AppToolbar/RoleSelection.tsx
martin-trajanovski: Yes it was causing because when we update the token user is updated as well in the user context and it was trying to execute useEffect again but that component is already removed because we re-render the view. It was firing error or warning in the console that `Can't perform a React state update on an unmounted component`.

peter-asztalos (approved):

PR #341, opened 2021-04-21
Swap 1360 as a developer i want to use shortcode as a filter

jekabs-karklins:
## Description

This PR aims to add a feature to be able to filter samples by proposal shortcode

## Motivation and Context

The PR implements a new Resolver for Samples type which returns back the associated proposal. This information is can be used for other things in the future as well. The thing to keep an eye on is that requesting information with resolvers could cause performance issues if there are many joints. In this case, however, it will be fast because there is only one join on the primary key.

## How Has This Been Tested

Added E2E tests

## Fixes

SWAP-1360

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/266


src/hooks/sample/useSamplesWithQuestionaryStatus.ts
```diff
@@ -0,0 +1,28 @@
...
+import { SampleWithQuestionaryStatus } from '../../models/Sample';
+
+export function useSamplesWithQuestionaryStatus(filter?: SamplesFilter) {
+  const [samples, setSamples] = useState<SampleWithQuestionaryStatus[]>([]);
+
+  const [samplesFilter, setSamplesFilter] = useState(filter);
+  const [loadingSamples, setLoadingSamples] = useState(false);
+  const api = useDataApi();
+
+  useEffect(() => {
```
martin-trajanovski, line 15: I suggest you add the check for unmounted component to avoid console warnings if component where the hook is used is destroyed before hook finished loading.

martin-trajanovski (approved):
Otherwise it looks good! 👀 💯

PR #360, opened 2021-05-06
As a beamline manager I should get all technical reviews and be able to reassign them to other instrument scientist

jekabs-karklins:
## Description

This PR aims to fix the user story where it is necessary to have a responsible person for the technical review.
image

## Motivation and Context

- This PR adds a new property for the instrument called "beamline manager".  
- The responsible person for the technical review is initially always set as a beamline manager.
- It is possible for a beamline manager, user officer, or responsible user to re-assign the task to someone else.
- The information about who did the review is visible to the SEP panel.

## How Has This Been Tested

Added E2E

## Fixes

SWAP-1518

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/278


src/components/instrument/CreateUpdateInstrument.tsx
```diff
@@ -31,13 +33,20 @@ const CreateUpdateInstrument: React.FC<CreateUpdateInstrumentProps> = ({
...
+  if (!usersData) {
+    return <UOLoader />;
+  }
   const initialValues = instrument
     ? instrument
     : {
         name: '',
         shortCode: '',
         description: '',
+        managerUserId: -1,
```
martin-trajanovski, line 49: Not sure why is this -1 as initial value

src/components/review/AssignTechnicalReview.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  if (!usersData) {
+    return null;
+  }
+
+  const userIdToUser = (userId: number | null) =>
+    usersData.users.find((user) => user.id === userId);
+
+  return (
+    <div>
```
martin-trajanovski, line 52: Do you need `<div>` here or it can just be `<>` react fragment?

src/components/review/AssignTechnicalReview.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const [selectedUser, setSelectedUser] = useState(
+    proposal.technicalReviewAssignee
+  );
+  const { usersData } = useUsersData({
+    userRole: UserRole.INSTRUMENT_SCIENTIST,
+  });
+
+  if (!usersData) {
+    return null;
```
martin-trajanovski, line 45: Isn't it better to return <UOLoader> instead of null or maybe not? I am not completley sure about the use of the component it is just like a question but maybe it already has loader in this screen.

src/components/review/AssignTechnicalReview.tsx
```diff
@@ -0,0 +1,117 @@
...
+import useDataApiWithFeedback from 'utils/useDataApiWithFeedback';
+import withConfirm, { WithConfirmType } from 'utils/withConfirm';
+
+type AssignTechnicalReviewProps = {
+  proposal: Proposal;
+  onProposalUpdated: (proposal: Proposal) => void;
+  confirm: WithConfirmType;
+};
+
+const getName = (user: BasicUserDetails | undefined) =>
```
martin-trajanovski, line 16: This `getName` function can be renamed to something like `getFullName` and moved to helper functions and reused in many places as we need full name in many components.

martin-trajanovski (approved):
Few small changes maybe but it looks good otherwise.

src/components/instrument/CreateUpdateInstrument.tsx
jekabs-karklins: When creating new instrument this must be set initialy to an invalid userId, because the creator has go specify it

src/components/review/AssignTechnicalReview.tsx
jekabs-karklins: If it's just `return` then it is equal to `return undefined`. React to rener needs either `React.element` or `null`

src/components/review/AssignTechnicalReview.tsx
jekabs-karklins: Yes, fixed +1

cypress/integration/calls.ts
```diff
@@ -223,15 +223,10 @@ context('Calls tests', () => {
...
-    cy.contains('Instruments').click();
-    cy.contains('Create').click();
-    cy.get('#name').type(instrumentAssignedToCall.name);
-    cy.get('#shortCode').type(instrumentAssignedToCall.shortCode);
-    cy.get('#description').type(instrumentAssignedToCall.description);
-    cy.get('[data-cy="submit"]').click();
-
-    cy.notification({ variant: 'success', text: 'successfully' });
+    cy.contains('People').click();
+    cy.addScientistRoleToUser('Carl');
```
JEThompson97, line 227: Very minor, but would be good to refactor 'Carl' into a variable, like you've done in `scheduler.ts`:
```ts
const scientist = 'Carl';
cy.addScientistRoleToUser(scientist);
```

src/utils/user.ts
```diff
@@ -0,0 +1,5 @@
+import { User } from 'generated/sdk';
+
+export const getFullUserName = (
```
JEThompson97, line 3: Nice refactor 👍

src/components/review/ProposalReviewContent.tsx
```diff
@@ -74,21 +90,59 @@ const ProposalReviewContent: React.FC<ProposalReviewContentProps> = ({
...
   );
 
+  const assignAnotherReviewerView = (proposal: Proposal) => {
+    if (proposal.technicalReview?.submitted) {
+      return null;
+    }
+
+    return (
+      <Paper elevation={1} className={classes.reassignContainer}>
+        <Typography variant="h6" gutterBottom>
```
JEThompson97, line 100: The use of a `h6` header here will cause the page to skip heading levels, which is bad for accessibility.
However, I'm working on [an overhaul of the heading levels](https://github.com/UserOfficeProject/stfc-user-office-project/issues/117) throughout the app, so I can pick this up when I make my PR for that.
I just mention for awareness.

JEThompson97 (approved):
Code looks good as far as I can tell 🙂

PR #361, opened 2021-05-06
Swap 1522 all sep members can see all proposals

martin-trajanovski:
## Description

SEP member is now able to see and review all the proposals on that SEP.

## How Has This Been Tested

manual tests
e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1522

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/280


src/components/review/ProposalTableReviewer.tsx
```diff
@@ -49,6 +53,13 @@ const getFilterStatus = (selected: string | ReviewStatus) =>
     ? ReviewStatus.DRAFT
     : undefined; // if the selected status is not a valid status assume we want to see everything
 
+const getFilterReviewer = (selected: string | ReviewerFilter) =>
```
jekabs-karklins, line 56: This is a little bit hard to read, maybe this logic can be simplified?

jekabs-karklins (approved):
Overall looks good

src/components/review/ProposalTableReviewer.tsx
martin-trajanovski: I just used the same logic as the previous `getFilterStatus` function but I agree maybe we should simplify both of them. I will try to do it.

PR #374, opened 2021-05-18
Fixing styling issues and adding sample shipments views

martin-trajanovski:
## Description

1. Added sample shipments table and fixed some bugs around it.
2. Fixed some styling issues around the page wrappers to expand to the screen width.

## How Has This Been Tested

1. e2e tests
2. manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1477

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/290

### DIFF
--- a/cypress/cypress.json
+++ b/cypress/cypress.json
@@ -3,6 +3,5 @@
   "baseUrl": "http://proxy:80",
   "screenshotsFolder": "/tmp/screenshots",
   "retries": 1,
-  "viewportWidth": 1300,
-  "viewportHeight": 1200
+  "scrollBehavior": "center"
 }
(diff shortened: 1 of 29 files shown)


jekabs-karklins:
Not sure, but think the extended view makes it harder to comprehend the view in some pages
image

Also the long title text looks too wide
image

I believe in some places it is good to utilize the full width but in many places it looks too stretched out.

martin-trajanovski:
> Not sure, but think the extended view makes it harder to comprehend the view in some pages
> image
> 
> Also the long title input looks too wide
> image
> 
> I believe in some places it is good to utilize the full width but in many places it looks too stretched out.

@jekabs-karklins I totally agree but not sure if only the ones that you mentioned are looking a bit weird or there are many more views that should be limited in width on very wide screens. I think we should go through all of them and decide.

jekabs-karklins (approved):

PR #377, opened 2021-05-21
feat: Swap 1570 as a user i want to be able to register a visit

jekabs-karklins:
## Description

As a user I want to be able to register a visit

## Motivation and Context

When user has a proposal accepted, he should be able to register for a visit. 
The registration form is based on templating system, that user officer can design.

## Fixes

SWAP-1517

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/292


cypress/integration/visitations.ts
```diff
@@ -0,0 +1,116 @@
+import faker from 'faker';
+
+faker.seed(1);
```
martin-trajanovski, line 3: What is this used for. I haven't used this seed before just wondering what it is doing.

cypress/integration/visitations.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  it('Should be able to delete visitation', () => {
+    cy.login('user');
+
+    cy.contains('My visitations').click();
+
+    cy.contains(proposalTitle)
+      .parent()
+      .get("[title='Delete visitation']")
+      .first()
```
martin-trajanovski, line 107: I think you don't need this .first() here because you are targeting the element by proposal title anyway.

src/components/common/FormikDropdown.tsx
```diff
@@ -106,29 +106,31 @@ const FormikDropdown: React.FC<
...
-      fullWidth
-      required={required}
-      disabled={disabled}
-      InputProps={InputProps}
-      {...props}
-    >
-      {children}
-      {menuItems}
-    </Field>
+    <>
```
martin-trajanovski, line 109: Not sure why this is wrapped inside an empty react fragment

src/components/questionary/Questionary.tsx
```diff
@@ -28,7 +28,7 @@ function Questionary({
   handleReset,
   displayElementFactory,
 }: QuestionaryProps) {
-  const isTabletOrMobile = useMediaQuery('(max-width: 1224px)');
+  const isTabletOrMobile = useMediaQuery('(max-width: 500px)');
```
martin-trajanovski, line 31: Like this maybe it should be called only `isMobile` instead of `isTabletOrMobile` because this resolution lower than 500 is not for tablets.

src/hooks/visitation/useVisitation.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export function useVisitation(visitationId: number) {
+  const [visitation, setVisitation] = useState<
+    GetVisitationQuery['visitation'] | null
+  >(null);
+
+  const api = useDataApi();
+
+  useEffect(() => {
+    api()
```
martin-trajanovski, line 14: `Unmounted` trick here to prevent some console warnings

src/hooks/visitation/useVisitations.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function useVisitations(filter?: VisitationsFilter) {
+  const [visitationsFilter, setVisitationsFilter] = useState(filter);
+  const [visitations, setVisitations] = useState<VisitationBasic[]>([]);
+  const [loadingVisitations, setLoadingVisitations] = useState(true);
+
+  const api = useDataApi();
+
+  useEffect(() => {
+    setLoadingVisitations(true);
```
martin-trajanovski, line 15: `Unmounted` here as well

martin-trajanovski (changes requested):

cypress/integration/visitations.ts
jekabs-karklins: It uses the same random values for the test depending on the seed given.
It is here to avoid false positives when running E2E

martin-trajanovski (approved):

PR #383, opened 2021-05-25
feat(proposals): Improve assigning and removing proposals to/from ins…

martin-trajanovski:
## Description

Improve assigning/removing proposal to/from instrument so we should be able to remove multiple proposals from an instrument at once.

## How Has This Been Tested

e2e tests
manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1434

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/302


src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -699,10 +670,10 @@ const ProposalTableOfficer: React.FC<ProposalTableOfficerProps> = ({
             position: 'toolbarOnSelect',
           },
           {
-            icon: AddScienceIcon.bind(null, {
-              'data-cy': 'assign-proposals-to-instrument',
+            icon: ScienceIconComponent.bind(null, {
+              'data-cy': 'assign-remove-instrument',
             }),
-            tooltip: 'Assign proposals to instrument',
+            tooltip: 'Assign/Remove instrument',
```
jekabs-karklins, line 676: It used to say 'Assign proposals', it now says 'Assign instrument' Is that correct?

jekabs-karklins (approved):

src/components/proposal/ProposalTableOfficer.tsx
martin-trajanovski: Yes it is a bit tricky to choose the right words. But I decided to go this way because when you select proposals it is obviously that you want to do something with them and that's why I didn't mentioned proposals in the tooltip. So the button is to assign instrument to proposal or remove instrument from proposal but Just Assign/Remove instrument is enough I think.

PR #384, opened 2021-05-25
120 win images built in gh actions

TCMeldrum:
Please see here for relevant detail: https://github.com/UserOfficeProject/user-office-backend/pull/303


martin-trajanovski (approved):

mr-field (approved):

TCMeldrum:
Does anyone know why the SEP tests might be failing?  I have re-run them one and got the same result, looks like there timing out?

martin-trajanovski:
> Does anyone know why the SEP tests might be failing? I have re-run them one and got the same result, looks like there timing out?

If you open the failing checks and see in the top right corner there is one button called `Artifacts` you can download and see some screenshots from failing tests. Maybe that way you will have better overview why they are failing. Do you try to run the tests locally to see if they pass?

TCMeldrum:
My changes don't change the frontend end just change our dockerfiles and creates a new github action. I don't see why that would change the outcome of the tests. I can try to see if they pass locally.

TCMeldrum:
Re ran the tests and they passed not sure what was happening

PR #397, opened 2021-05-31
patch(questionary): Improving naming and typechecking

jekabs-karklins:
## Description

This PR aims to address the problem where events dispatched in context were having payloads of type `any`. This was allowing to pass arbitrary data and was a potential cause for bugs. 

Now depending on event type, payload format is fixed.

There are many changed files, but the changes are monotonous 

## Fixes

SWAP-1600 Improving naming and typechecking

## Changes

Questionary

## Depends on

N/A


src/components/proposal/ProposalContainer.tsx
```diff
@@ -38,30 +37,30 @@ const proposalReducer = (
...
+    case 'PROPOSAL_LOADED':
+      const proposal = action.proposal;
       draftState.isDirty = false;
       draftState.questionaryId = proposal.questionaryId;
       draftState.proposal = proposal;
       draftState.steps = proposal.questionary?.steps || [];
       draftState.templateId = proposal.questionary?.templateId || 0;
       break;
-    case EventType.PROPOSAL_MODIFIED:
+    case 'PROPOSAL_MODIFIED':
```
martin-trajanovski, line 49: Not sure why instead of the enums it is better to use strings to compare the action type. I think it was better to use the enum to prevent typos and some other mistakes.

PR #403, opened 2021-06-03
81 update proposal editor interface to improve clarity

Ziggiyzoo:
## Description

Closes UserOfficeProject/stfc-user-office-project#81

Clarity improvements were needed with regards to proposals being edited. Improvements to the step-index logic and actions icons in the dashboard were required after adding the ability for proposals to be edited after submission.

## Motivation and Context
Proposals after being submitted and having status `EDITABLE_SUBMITTED`

- Proposals would continue to have the "View" Icon rather then edit icon
- Proposals would open at the Review page, rather than on the proposal or any other topic pages.
- Proposals no with the `EDITABLE_SUBITTED` status would show as editable when they should be read-only.

## How Has This Been Tested

Using standard call:

- Created proposal and submitted.
   - Proposal remains in a read-only state and cannot be edited.
- Created proposal and left as saved.
   - Proposal will open at the last unfinished topic, or if they are all complete, will open on the review page.

Using Editable-Submitted call:

- Created proposal and submitted.
   - Proposal opens on the first topic and is editable, allowing the user to follow the same steps of `Save & Continue` as with creating the proposal.
- Created proposal and left as saved.
   - Proposal will open at the last unfinished topic, or if they are all complete, will open on the review page. 

A repeat of these tests with a modified call template yielded the expected results.

## Changes

Addition and alteration of some `if` statement operations.


src/components/proposal/ProposalTable.tsx
```diff
@@ -149,11 +149,26 @@ const ProposalTable = ({
...
+                  if (rowData.status == 'EDITABLE_SUBMITTED') {
+                    return false;
+                  } else {
+                    return true;
+                  }
+                }
+              } else if (!isCallActive) {
+                return true;
+              }
+            }
```
mr-field, line 167: I think this could be simplified to:
```
const readOnly = !isCallActive || (rowData.submitted && rowData.status !== 'EDITABLE_SUBMITTED');
```

mr-field (commented):
Looks good, just a suggestion for a minor improvement

src/components/proposal/ProposalTable.tsx
```diff
@@ -149,7 +149,9 @@ const ProposalTable = ({
         actions={[
           (rowData) => {
             const isCallActive = rowData.call?.isActive ?? true;
-            const readOnly = !isCallActive || rowData.submitted;
+            const readOnly =
+              !isCallActive ||
+              (rowData.submitted && rowData.status == 'EDITABLE_SUBMITTED');
```
TCMeldrum, line 154: Is the logic here correct? If a proposal has the status EDITABLE_SUBMITTED and is submitted then the right-hand side of the or will be true so the proposal will readOnly. Conversly if the call is still active, the proposal is submitted but is not EDITABLE_SUBMITTED then it will not be readOnly? I think the == be !==.

mr-field: Yes, I think this should be `!==`. Would you be OK to make the change?

TCMeldrum: Yeah will do

Ziggiyzoo: Oh yes.. It should be !== I will sort that now! Are there any other changes required? I'm looking into the tests at the moment.

TCMeldrum: I have made this change so the proposal test is now passing but the templet tests are still failing. Seems like when you have submitted a proposal and you are reviewing it you can't seem to click on a button.

Ziggiyzoo: So that button is appearing as read-only? Where is that button? I may have completely missed it before

TCMeldrum: This button is when reviewing a submitting a proposal, it looks at this page in tests to check the proposal has been submitted.

src/components/proposal/ProposalContainer.tsx
```diff
@@ -77,11 +77,18 @@ const proposalReducer = (
...
+  }
+}
+
+function isReadOnly(proposalToCheck: ProposalSubsetSubmission) {
   if (
-    proposal.status != null &&
-    (proposal.status.shortCode.toString() === 'DRAFT' ||
-      proposal.status.shortCode.toString() === 'EDITABLE_SUBMITTED')
+    !proposalToCheck.submitted ||
+    GetProposalStatus(proposalToCheck) === 'EDITABLE_SUBMITTED'
```
TCMeldrum, line 91: After running the tests locally it seems like this is the line that fails the final test. By adding 'DRAFT' back as an option in here it seems to fix it. However the logic does seem correct, it is read-only if it hasn't been submitted or if it is editable after submission.

TCMeldrum: After some investigation I think this read-only function is just for viewing a submitted proposal not for editing proposal. I think that proposals that are editable should not get to this code at least while the call is active. So I think this change needs to be reverted but would like some clarification if there is someone some more familiar with this section of the frontend.

jekabs-karklins: I looked into it. It looks like the read-only state is set properly, however, for the ProposalSummary whole view was disabled, and the test was trying to click the view RichText content button but was unable to because all mouse pointers were disabled. I made a PR to address the problem. 
https://github.com/UserOfficeProject/user-office-frontend/pull/421

TCMeldrum: Cool thanks for the explanation, I was a bit confused by this.

simonfernandes: When I ran this test locally, the proposal used in the test has a submitted value of true, but it somehow has a status of 1 (draft). Which would explain why adding the OR draft check back in does get the test passing again. 

I think commit 02002c6 is still necessary, as this read-only logic also seems to be used in the questionary (e.g. when you create a new proposal, whether the fields are disabled or not). So if it's in editable submitted status read only should be false to enable users to edit proposals in this state (which there aren't currently any tests for).

simonfernandes (approved):
It looks fine to me. Obviously I made that last commit though so it would still be good to get someone else's check on everything still

TCMeldrum:
https://github.com/UserOfficeProject/user-office-frontend/pull/403 Should be merged into this branch before this is merged into develop. With 02002c6 it would be helpful for me if you could explain why we need the DRAFT state as well as the EDITABLE_SUBMITTED state?

simonfernandes:
Regarding 02002c6, the read only function seems to also affect whether the fields in a proposal questionnaire are disabled or not (when you click edit on an existing proposal or click new proposal). 

Setting read only to false when it's in editable submitted status allows the user to actually make modifications to their proposal so is needed.

Setting read only to false when it's in draft status probably isn't strictly necessary and could be removed (because we already set it to false when a proposal is not submitted, and in theory proposals that are submitted shouldn't be in draft status). But in this case there's a test proposal that is submitted and is someohow still in draft status, so the draft check says that the proposal isn't read only and allows the rich text button to be enabled. Jekab's fix is a better way to get it passing because the proposal in the test should be read only

TCMeldrum:
Cool thanks!

jekabs-karklins (approved):

src/components/proposal/ProposalSummary.tsx
```diff
@@ -74,12 +70,20 @@ function ProposalReview({ readonly, confirm }: ProposalSummaryProps) {
     !allStepsComplete ||
     proposal.submitted;
 
+  function SubmitButtonMessage() {
+    if (
+      proposal.status != null &&
+      proposal.status.shortCode.toString() == 'EDITABLE_SUBMITTED'
```
simonfernandes, line 76: I realised that this will never be shown, because when a user modifies a proposal in the `EDITABLE_SUBMITTED` state, the proposal is saved, not re-submitted.

I think we instead want to somehow look ahead to the upcoming proposal workflow statuses and show this if one of them is `EDITABLE_SUBMITTED`.

simonfernandes (changes requested):

simonfernandes:
I've had a go at this and I'd be keen to know if there is a better way to go about it. I did originally create a separate hook, but then found that you can't conditionally use them. I wasn't quite sure on when the best time is to do the checks. I have tried delaying the submit button load, binding the check to the click etc., but went with this in the end. This shows a loading spinner on the page until the submit message is known, then once the page re-renders it behaves as it normally would.

simonfernandes:
Hi @UserOfficeProject/swap, just a friendly reminder that this is waiting on an ESS review at your earliest convenience 😊

martin-trajanovski:
> Hi @UserOfficeProject/swap, just a friendly reminder that this is waiting on an ESS review at your earliest convenience 😊

Isn't this approved from @jekabs-karklins? I see he has approved it few days ago.

simonfernandes:
@martin-trajanovski It is, but I have since made a couple of changes that I am not entirely confident in, so if someone wouldn't mind having another look that would be great

martin-trajanovski (approved):
For me it looks good but lets see how it works in action later on.

PR #412, opened 2021-06-09
Add STFC theme

simonfernandes:
## Description

This refactors the theme so that it now reads colours from variables in the Settings table of the database. If there are colours defined in any of the theme variables, the Material UI theme is updated as well as the CSS variables. If no colours are defined in the database, the default Material UI colours are used.

This also allows for an optional header image. The filename can be specified in the Settings table and it will be used as an import by looking for the file in the images directory.

## How Has This Been Tested

Tested with empty values in all of the theme variables (default colours are used), and with.

## Fixes

Part of UserOfficeProject/stfc-user-office-project#69 

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/349 (backend branch)

None

## Screenshots

87cfea0bb2174a839cfccab5f358492f
34be5b85650643f5959ef7dff245a7b1
d1367dbdf06541798cec6518b6003733


src/components/AppToolbar/AppToolbar.tsx
```diff
@@ -69,6 +77,17 @@ const AppToolbar: React.FC<AppToolbarProps> = ({ open, handleDrawerOpen }) => {
...
+    logoContainer: {
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center',
+      marginRight: 24,
+      ...theme.mixins.toolbar,
+    },
+    logo: {
+      // Ensures min 1/4 of the height is padding
+      maxHeight: '71.4%',
```
simonfernandes, line 89: This is a bit ugly, but I struggled to find another way to do this. The space around the image needs to be 20% of the image's height per UKRI rules on logo use. When using padding with percentages, it's calculated from the width of the available space rather than height. I wanted to get the height of the toolbar and calculate it based on that, e.g.

``padding: `calc(${theme.mixins.toolbar.height} * 0.2)` ``

but the toolbar height just kept coming back as undefined.

TCMeldrum (commented):
Looks good, had a look on my local set up and it looks good! Eslint gave a few warnings do we need to fix those?

simonfernandes:
> Looks good, had a look on my local set up and it looks good! Eslint gave a few warnings do we need to fix those?

Is this running `npm run lint`? It looks like that didn't come back with anything for me. I didn't bypass the linter in any of my commits so I'm hoping I haven't introduced anything

src/themes/theme.ts
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import { Theme } from '@material-ui/core/styles';
+
```
TickleThePanda, line 3: I think this could include a function `getHeaderLogo` which keeps some of the specific code here. This could include any information about adding additional padding to the logo rather than including facility specific code elsewhere.

src/components/AppToolbar/AppToolbar.tsx
```diff
@@ -15,6 +16,13 @@ import { UserContext } from 'context/UserContextProvider';
 
 import AccountActionButton from './AccountActionButton';
 
+let headerLogo: string | undefined;
+switch (process.env.REACT_APP_AUTH_PROVIDER) {
```
TickleThePanda, line 20: I'm not sure if we should be using `REACT_APP_AUTH_PROVIDER` as the environment variable. Possibly `USER_OFFICE_FRONTEND_THEME`.

TCMeldrum:
OK cool, it was with the docker container itself but running npm run lint gives me nothing either. Sorry still trying to figure this all out.

TCMeldrum (approved):

src/index.tsx
```diff
@@ -5,6 +5,13 @@ import './index.css';
 import App from './components/App';
 import * as serviceWorker from './serviceWorker';
 
+const org = process.env.REACT_APP_AUTH_PROVIDER;
+switch (org) {
+  case 'stfc':
+    require('./stfc-stylesheet.css');
```
TickleThePanda, line 11: Is there any way to move this logic to `theme.ts`?

TickleThePanda (changes requested):
Some comments about moving theme switching code to the `theme.ts` class and about the environment variable.

@UserOfficeProject/swap, are you happy with this specific code being switched like this? I can't particularly think of another way to do it here.

src/components/AppToolbar/AppToolbar.tsx
simonfernandes: I did question it at one point but then just moved forward with it! Something with theme in it would be more descriptive.

I think it has to include the [`REACT_APP` prefix](https://create-react-app.dev/docs/adding-custom-environment-variables/), so what about `REACT_APP_FRONTEND_THEME`?

src/index.tsx
simonfernandes: I couldn't quite get this working unfortunately. I had a function in the theme to get the styling, which returned a require import for the CSS. 

The main problem I found was that the import appeared unused, even though it did seem to be calling the function (and I think might have worked). It wouldn't compile though and there's no typescript eslint rule that I could find to override this. I'm also unsure if this would actually have worked when the get styling function returns undefined (in the case of the ESS).

src/index.tsx
```diff
@@ -1,10 +1,17 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-import './index.css';
+import './css/shared-stylesheet.css';
 import App from './components/App';
 import * as serviceWorker from './serviceWorker';
 
+const org = process.env.REACT_APP_THEME;
```
bolmsten, line 8: Can this be removed now? I see the same lines still in the theme.ts file

simonfernandes: Unfortunately not as this is to import the additional css file. In theme.ts, that is used to choose which theme to use.

simonfernandes:
Ready to review again with the changes to use colours stored in the database - I've updated the description and added a backend branch

martin-trajanovski (approved):

jekabs-karklins (approved):

PR #444, opened 2021-07-02
Adjust heading levels to be sequential

JEThompson97:
## Description

This adjusts the headings throughout the app to make heading-levels **more sequential and avoid gaps** between levels.

Previous to this change, many of the headings jumped to `h6` by default (among other issues).

##### Example difference:
| Before | This PR |
|--------|--------|
|image|image|

#### Modals
For modals, we decided on the design principle to restart heading-levels from `h1`. From our research, we determined there was no clear rule for the case of modals, but that treating them as the beginning of a new heading hierarchy was a justifiable option.

#### Re-use conflicts
There are instances of re-used components where one use of a components wants a given heading level, X, and other use wants a different level, Y. [These conflicts are documented here, in a **follow-up issue**](https://github.com/UserOfficeProject/stfc-user-office-project/issues/139). In these cases, the **highest** expected heading-level has been selected for this PR as an improvement, pending the follow-up issue being worked on. E.g. where `h2` and `h3` are both expected in different instances, `h2` has been used.

---

In addition, this PR **adds tooltips** to User Officer menu items.
tooltips

**_NOTE:_** this PR **only** covers areas of the app accessible with **STFC** config. However, these changes can be easily replicated to additional areas.

## Motivation and Context

This improves the accessibility of the app for screen-reader users
[Source](https://www.w3.org/WAI/tutorials/page-structure/headings/#heading-ranks)

## How Has This Been Tested
Manual testing

## Fixes
Fixes https://github.com/UserOfficeProject/stfc-user-office-project/issues/117

## Changes
- Changes heading level components (visuals unaffected)
- Adds tooltips for menu items


src/components/MenuItems.tsx
```diff
@@ -54,57 +54,69 @@ const SettingsMenuListItem = () => {
```
JEThompson97, line 54: **Note:**

GitHub isn't showing the changes I've made in this file very well.
**All I've done here is wrap tooltips around menu items.** And ensured the `Tooltip` text matches the `ListItemText`

src/components/questionary/Questionary.tsx
```diff
@@ -60,7 +60,7 @@ function Questionary({
   const classes = useStyles();
   const { state, dispatch } = useContext(QuestionaryContext);
   const isUserOfficer = useCheckAccess([UserRole.USER_OFFICER]);
-  const titleRef = useRef<HTMLElement | null>(null);
+  const titleRef = useRef<HTMLHeadingElement | null>(null);
```
JEThompson97, line 63: I had to make this change to allow me to add the `component` attribute to the Typography component this is used in.

I'm not sure if this has any wider implications? It didn't seem to in manual testing...

src/components/MenuItems.tsx
ACLay: There's a setting that allows you to ignore whitespace changes when reviewing a PR. It's quite handy for when blocks of text have their indentation changed: 
image

ACLay (approved):
I think the main PR body needs a slight edit. It currently says "In these cases, the highest expected heading-level has been selected for this PR as an improvement pending a full solution", which reads to me as though a follow up PR is expected, but it also contains a "Fixes <issue>" so merging this in would automatically close the issue. 

Otherwise, I think the changes look okay.

src/components/MenuItems.tsx
JEThompson97: Oh brilliant, that's great! Cheers for pointing that out

JEThompson97:
@ACLay 
> which reads to me as though a follow up PR is expected

Ah, something I didn't make clear in the PR body is that I've made a follow up issue to address the move complicated heading level issues (which is what I mean by 'a full solution'): https://github.com/UserOfficeProject/stfc-user-office-project/issues/139

I'll make this clearer 🙂

src/components/eventLog/EventLogList.tsx
```diff
@@ -44,7 +45,11 @@ const EventLogList: React.FC<EventLogListProps> = ({
     <div data-cy="event-logs-table">
       <MaterialTable
         icons={tableIcons}
-        title={'Event logs'}
+        title={
+          <Typography variant="h6" component="h2">
+            {'Event logs'}
```
jekabs-karklins, line 50: I believe we can simply type `Event logs` instead of `{'Event logs'}` here because it is a string literal within a JSX

src/components/institution/InstitutionTable.tsx
```diff
@@ -69,7 +70,11 @@ const InstitutionPage: React.FC = () => {
         setData={setInstitutions}
         createModal={createModal}
         icons={tableIcons}
-        title={'Institutions'}
+        title={
+          <Typography variant="h6" component="h2">
+            {'Institutions'}
```
jekabs-karklins, line 75: And same here `Instructions` , there are several place like this

jekabs-karklins (approved):

PR #453, opened 2021-07-08
Improve reference number format UX

RasmiaKulan:
Closes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/110

### Changes made
1. **Added validation to check if the inputted reference number format is correct.**

updated-validation

2. **Added a modal to display information about what the reference number format consists of and what the system expects to be a valid format.**
final-modal


src/components/call/CallGeneralInfo.tsx
```diff
@@ -53,6 +74,62 @@ const CallGeneralInfo: React.FC<{
...
+      if (input.match(regExp)) {
+      } else if (!input.match(suffix)) {
+        error = 'Invalid parameter.';
+      } else if (!input.match(prefix)) {
+        error = 'Invalid parameter.';
+      }
+    }
+
+    return error;
+  }
```
JEThompson97, line 92: I notice you've replaced having `Invalid Suffix` and `Invalid Prefix` with just `Invalid parameter` for both.
Was this intentional? If so, I think the code can be simplified a little:
```suggestion
  function validateRefnumber(input: string) {
    const regExp = /((([a-z|\d]+)){)({digits:[1-9]+})/;
    if (input && !input.match(regExp)) {
      return 'Invalid parameter.';
    }
  }
```
Please check this works before committing 😅

src/components/call/CallGeneralInfo.tsx
```diff
@@ -90,9 +167,77 @@ const CallGeneralInfo: React.FC<{
...
+                      The<strong> prefix</strong> can contain alphanumeric
+                      characters and is what all generated reference numbers
+                      will begin with. For example, <code>21a</code>.<br></br>
+                      <br></br>
+                      The <strong> digits parameter</strong> is a numerical
+                      value that determines how many digits a proposal&apos;s
+                      sequence number is, including padding.It is written as
+                      <code>{'{digits:x}'}</code>, where x is a the value of the
+                      number. For example, if parameter is 6 (
+                      <code>{'{digits:6}'}</code>),the first proposal will be
```
JEThompson97, line 201:
```suggestion
                      <code>{'{digits:6}'}</code>), the first proposal will be
```
image

src/components/call/CallGeneralInfo.tsx
```diff
@@ -90,9 +167,77 @@ const CallGeneralInfo: React.FC<{
...
+                      numbered 000001, the second 000002, and so on.
```
JEThompson97, line 202: Big fan of the Oxford comma 👍😁

JEThompson97 (commented):
A few comments from a first glance 🙂

Visually and UX wise, I think this has turned out really nice!
I like especially like how you've used **bold** and `code sections` to make the help text easier to read. Great stuff 👍

src/components/call/CallGeneralInfo.tsx
RasmiaKulan: @EdwardHaynes Thank you so much for having a look at this. 😊
I misunderstood what @simonfernandes mentioned before about the error messages and didn't update the code properly. He suggested displaying `Invalid prefix` and `Invalid parameter` for the invalid suffix. I need to make those changes.

src/components/call/CallGeneralInfo.tsx
```diff
@@ -53,6 +74,62 @@ const CallGeneralInfo: React.FC<{
     }
   }, [startCall, endCall, formik]);
 
+  function validateRefnumber(input: string) {
+    let error;
+    const regExp = /((([a-z|\d]+)){)({digits:[1-9]+})/;
+    const prefix = /((([a-z|\d]+)){)/;
+    const suffix = /({digits:[1-9]+})/;
```
DChambersSTFC, line 81: Although those these aren't too complex, it might be helpful for future work to have a few comments explaining what these regexes are looking for :)

src/components/call/CallGeneralInfo.tsx
```diff
@@ -53,6 +74,62 @@ const CallGeneralInfo: React.FC<{
...
+      if (input.match(regExp)) {
+      } else if (!input.match(prefix)) {
+        error = 'Invalid prefix.';
+      } else if (!input.match(suffix)) {
+        error = 'Invalid parameter.';
+      }
+    }
+
+    return error;
+  }
```
JEThompson97, line 92: Thanks for clarifying about the error messages 🙂.
Perhaps we could simplify by removing the `regExp` bit? (it seems the sub-regexs do all the necessary checks).
I've also fleshed out the variable names for a bit more clarity (but I'll leave the final decision with you):
```suggestion
  function validateRefnumber(input: string) {
    let errorMessage;
    const prefixRegex = /((([a-z|\d]+)){)/;
    const parameterRegex = /({digits:[1-9]+})/;
    if (input) {
      if (!input.match(prefixRegex)) {
        errorMessage = 'Invalid prefix.';
      } else if (!input.match(parameterRegex)) {
        errorMessage = 'Invalid parameter.';
      }
    }
    return errorMessage;
  }
```
**Again, please check it works before committing** 😁

src/components/call/CallGeneralInfo.tsx
```diff
@@ -90,9 +167,77 @@ const CallGeneralInfo: React.FC<{
...
+                </IconButton>
+                <Dialog
+                  onClose={handleClose}
+                  aria-labelledby="customized-dialog-title"
+                  open={open}
+                  color="primary"
+                >
+                  <DialogContent dividers>
+                    <Typography gutterBottom color="inherit" variant="body1">
+                      A reference number format determines how reference numbers
```
simonfernandes, line 188: You could put this text in its own variable to keep the render() method a little bit tidier. Similarly you could also pull out the table (which is quite verbose) into a component, though I'm not sure how necessary that is

src/components/call/CallGeneralInfo.tsx
simonfernandes: I think a main regex might still be necessary @JEThompson97, or the user could theoretically have some whitespace between the prefix and parameter and the check would still pass. I like the naming though, that seems clearer 😊

Some possible improvements to the regex part: 

- Might be a good idea to include the `^` and `$` characters for the start and end of string to the main regex. This will avoid the user adding whitespace before and after an otherwise valid format.
- It looks like in the full regular expression you are defining groups in there (the brackets), so I think you could get rid of the prefix and suffix regexes and instead check the indices of the match (https://javascript.info/regexp-groups#parentheses-contents-in-the-match) to check the prefix and parameter groups (see example below). This will be more efficient because you're only executing one match. There is a nicer way to refer to the groups by given names instead of indices, but I know we struggled to get this working when we had a look @RasmiaKulan. 

- Instead of doing nothing when the input matches the regex, it would be better practice (i.e. more explicit and easier to read) if you invert the condition and then do the additional checks 
  ```javascript
  const match = input.match(regExp);
  if (!match[0]) {
    if (!match[1]) {
      error = 'Invalid prefix';
    } else if (!match[2]) {
      error = 'Invalid parameter';
    } else {
      error = 'Invalid format';
  }
  ```

- Have an else condition that is just 'Invalid format' (as above), for when the input doesn't match the regex, but does match the prefix and parameter. I think this input be an example of this: `192      {digits:4}` (whitespace between the capture groups)
- Capitalise the N so that it's `validateRefNumber`

simonfernandes (commented):
As Edward said, I think it looks great and the additional error messages are a nice touch! 👍 I've added some suggestions for regex improvements but nothing major 😊

jekabs-karklins (approved):

simonfernandes (approved):

PR #461, opened 2021-07-12
fix(tepmlates): Improving archived templates queries

martin-trajanovski:
## Description

Improved `getTemplates` query so that it returns archived template only if `templateIds` is passed to show it in call/other template where it is already used but after that it is archived. Previously if template gets archived and it is already used in a call or another template empty dropdown was shown because it wasn't returned from the query. We were returning only unarchived templates.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1412

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/352


src/graphql/call/fragment.call.graphql
```diff
@@ -33,5 +33,10 @@ fragment call on Call {
     name
     description
   }
+  template {
```
jekabs-karklins, line 36: I would suggest to not adding values retrieved from resolvers to the fragment and keep fragment lightweight.

martin-trajanovski: @jekabs-karklins I see for example `instruments` was added like 9 months ago in this fragment from resolvers. Not sure if I should clean everything up or how do you suggest to do it?

jekabs-karklins: I think yes, we should keep the fragments lightweight, so if we can move out more then the better. But I guess not with this PR.

jekabs-karklins (approved):

PR #467, opened 2021-07-13
146 change user selection for proposals

TCMeldrum:
Closes https://github.com/UserOfficeProject/stfc-user-office-project/issues/146 and https://github.com/UserOfficeProject/stfc-user-office-project/issues/152

## Description

Replaced peopleTable for users with proposalPeopleTable so that they cannot see the full user list. Now they can only see users that they have previously collaborated with as well as inviting existing users by email. 

I based the new component on the people table so it has some of the same code and structure.

image

## Motivation and Context

This solves this GDPR issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/146

## How Has This Been Tested

For the cypress tests, I have modified them to work on the new table. I then create a dummy proposal then log in as a user officer and edit the proposal from there to render a peopleTable so both tables are tested. 

## Changes

Created new component proposalsPeopleTable.tsx that is rendered when a user is creating a proposal. With this component, users can see users subset of the users that they have previously collaborated with or add an existing user by email. 

The new preivousCobaloraters query finds a user's most recent collaborators (the collaborators from the most recent proposal they worked on) as well as their 10 most collaborated with users.

The email box has some validation on it giving different error messages e.g. User is already on the proposal. It also renders a more detailed message if the email queried does not exist as this within our existing system as there are privacy settings possibly stopping a user from being found. This message may be something we could make site-specific if needed. 

We have also added some logic about who appears in add principal investigator/co-proposer lists based on if they are on the project already.

- If a user is on the proposal already as a principle investigator they will not appear in the lists
- If a user a co-proposer they will not be able to be added a co-proposer. 
- If a user is a co-proposer they can be 'elevated' to the principal investigator, this will remove them as a co-proposer and remove the current principal investigator from the proposal
- If a user removes someone as principal investigator they will become a co-proposer.
- A user cannot remove themselves as a co-proposer

There is now a partipantModel that renders peopleTable or proposalPeopleTable based on if the current userRole, if it, not a user officer, and if a new argument called participant (which is set in proposalParticipant(s) components) is set it renders the proposalPeopleTable else it renders peopleTable.

Created 2 new gql querys: getPreviousColaborators and getBasicUserDeatailsByEmail. 

## Possible Inprovments

Users that have been found by email could have some permeance within the list as currently they are lost when proposalPeopleTable is closed

This is my first time properly coding with react so if there is anything that can be improved around that please let me know. 

I am not sure where else this table might be used or if I have inadvertently placed it somewhere I should not.

## Depends on

Backend PR: https://github.com/UserOfficeProject/user-office-backend/pull/354

Edit: Updated the PI/CoPro logic to complete issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/152


src/components/user/ProposalsPeopleTable.tsx
```diff
@@ -0,0 +1,515 @@
...
+            if (props.selection)
+              setSelectedParticipants(
+                selectedParticipants.concat([userDetails])
+              );
+
+            //This is the recommend way to refresh a table but it is unsupported in typescript and may not be supported anytime soon
+            //See: Refresh Data Example https://material-table.com/#/docs/features/remote-data
+            //and see: https://github.com/mbrn/material-table/issues/1752
+            // @ts-ignore
+            tableRef.current.onQueryChange();
```
TCMeldrum, line 423: I could not find a better solution for this. If anyone else can I would happy to change it.

AlbertoPaulPopescu:
Sorry if this is a problem on my end only but I receive this error while trying to add a co-proposer. I tried to look up this issue but am not sure what the solution is.
image

TCMeldrum:
Ah yes, I tested this not with the STFC datasource as it was easier to see users in the database. I think I have to fix that now. Let me know if it keeps happening @AlbertoPaulPopescu.

AlbertoPaulPopescu:
> Ah yes, I tested this not with the STFC datasource as it was easier to see users in the database. I think I have to fix that now. Let me know if it keeps happening @AlbertoPaulPopescu.

I'm checking develop now and seems to be working. I'll check back your branch and update you on that. Also, is there an email address I can use to search for a co-proposer? @TCMeldrum

TCMeldrum:
@AlbertoPaulPopescu I am not sure what emails are of the user in the database when you can use with the STFC config sorry

TickleThePanda (approved):

src/components/user/ProposalsPeopleTable.tsx
martin-trajanovski: @TCMeldrum I think you can use something like this https://github.com/UserOfficeProject/user-office-frontend/blob/develop/src/components/user/PeopleTable.tsx#L244 if I understand the need correctly it is just to refresh the table data after you submit the form.

src/components/user/ProposalsPeopleTable.tsx
```diff
@@ -0,0 +1,514 @@
...
+        }
+      }}
+    >
+      {() => (
+        <div data-cy="co-proposers" className={classes.tableWrapper}>
+          <MaterialTable
+            tableRef={tableRef}
+            icons={tableIcons}
+            title={
+              <Typography variant="h6" component="h2">
```
JEThompson97, line 443: Is this the "Add Co-Proposers" title?

If so, I think the header component might want to be `h1` since it is the first heading of the modal (we decided as part of [this issue](https://github.com/UserOfficeProject/stfc-user-office-project/issues/117) to treat modals as the start of a new heading hierarchy.)
```suggestion
              <Typography variant="h6" component="h1">
```

JEThompson97 (commented):
Nice work with this one 🙂. I like the comments you've added around your new code; makes it easier to figure out what's going on.

Just a small comment from me

src/components/user/ProposalsPeopleTable.tsx
TCMeldrum: Sorry forgot to reply earlier, thanks for catching this!

src/components/proposal/ParticipantModal.tsx
```diff
@@ -27,6 +30,27 @@ function ParticipantModal(props: {
     ]);
   };
 
+  const someProps = {
```
bolmsten, line 33: renamne someProps to something else like userTableProps

bolmsten (approved):

PR #469, opened 2021-07-14
SWAP-1399-template-and-workflow-required-when-creating-call

martin-trajanovski:
## Description

Update validation for calls. Update yup package to match the version from validation package. Fix typescript errors after yup update to latest

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1399

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/357


src/components/questionary/questionaryComponents/SampleDeclaration/SampleDeclaratonDefinition.tsx
```diff
@@ -27,7 +26,7 @@ export const sampleDeclarationDefinition: QuestionaryComponentDefinition = {
   },
   createYupValidationSchema: (answer) => {
     const config = answer.config as SubtemplateConfig;
-    let schema = Yup.array().of(Yup.object<SampleWithQuestionary>());
```
jekabs-karklins, line 30: Do you know if Yup removed a possibility to type Array?

martin-trajanovski: I think it's not removed but I guess they have improved some types and it makes some problems with our generated types where we have those properties like __typename and so on and the easiest solution was to just use AnyObjectSchema for now. Maybe it can be resolved more elegant later but it needs some more investigation how to do it.

src/components/call/CreateUpdateCall.tsx
```diff
@@ -89,15 +89,13 @@ const CreateUpdateCall: React.FC<CreateUpdateCallProps> = ({ call, close }) => {
...
+          } = values as UpdateCallInput;
           if (call) {
             const data = await api('Call updated successfully!').updateCall({
               ...restValues,
               id: id,
-              templateId: templateId ? +templateId : null,
-              proposalWorkflowId: proposalWorkflowId
-                ? +proposalWorkflowId
-                : null,
+              templateId: +templateId,
```
jekabs-karklins, line 97: I don't think you need a + here anymore

src/components/call/CreateUpdateCall.tsx
```diff
@@ -89,15 +89,13 @@ const CreateUpdateCall: React.FC<CreateUpdateCallProps> = ({ call, close }) => {
...
+              proposalWorkflowId: +proposalWorkflowId,
```
jekabs-karklins, line 98: and here

src/components/call/CreateUpdateCall.tsx
```diff
@@ -106,10 +104,8 @@ const CreateUpdateCall: React.FC<CreateUpdateCallProps> = ({ call, close }) => {
           } else {
             const data = await api('Call created successfully!').createCall({
               ...restValues,
-              templateId: templateId ? +templateId : null,
-              proposalWorkflowId: proposalWorkflowId
-                ? +proposalWorkflowId
-                : null,
+              templateId: +templateId,
+              proposalWorkflowId: +proposalWorkflowId,
```
jekabs-karklins, line 108: same

src/components/questionary/questionaryComponents/FileUpload/FileUploadDefinition.tsx
```diff
@@ -25,7 +25,8 @@ export const fileUploadDefinition: QuestionaryComponentDefinition = {
     answerRenderer: FilesAnswerRenderer,
     questionRenderer: defaultRenderer.questionRenderer,
   },
-  createYupValidationSchema: () => Yup.array().of(Yup.string()),
+  createYupValidationSchema: () =>
+    Yup.array().of(Yup.object({ id: Yup.string().required() })),
```
jekabs-karklins, line 29: I think we added that file can also have a figure to the schema
`Yup.array().of(Yup.object({ id: Yup.string().required(), figure: Yup.string() }))`

jekabs-karklins (approved):

src/components/questionary/questionaryComponents/FileUpload/FileUploadDefinition.tsx
martin-trajanovski: This is temporary just to make tests pass and the next PR will make this validation schema come from the validation package and there it is already set up. But thanks for pointing it.

PR #482, opened 2021-07-23
SWAP-1435-improve-clonning-proposals

martin-trajanovski:
## Description

Improve cloning of proposals to be able to clone multiple proposals at once

## Motivation and Context

Previously cloning proposals was able only one by one now we can do multiple at once.

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1435

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/362


src/components/proposal/ProposalTableOfficer.tsx
```diff
@@ -179,7 +176,7 @@ const ProposalTableOfficer: React.FC<ProposalTableOfficerProps> = ({
             <Visibility />
           </IconButton>
         </Tooltip>
-        <Tooltip title="Clone proposal">
+        {/* <Tooltip title="Clone proposal">
```
jekabs-karklins, line 179: I think we should not leave commented code. All the changes are stored in  SVC.

jekabs-karklins (approved):

src/components/proposal/ProposalTableOfficer.tsx
martin-trajanovski: You are right I forgot this one 😅

PR #487, opened 2021-07-27
SWAP-1682-changing-statuses-improvement

martin-trajanovski:
## Description

Improve proposal status change with showing selected status for selected proposal/s if it's the same

## Motivation and Context

Previously if you select one or multiple proposals and try to change status you can not see the current status already selected in the dropdown. Now if the status for all selected proposal/s is the same it is shown.

## How Has This Been Tested

- manual tests
- e2e tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1682


jekabs-karklins (approved):
💯

PR #519, opened 2021-08-10
feat: implement searching the question drawer [Swap-1728]

jekabs-karklins:
## Description

Adding ability to search questions in the question drawer
image

## Motivation and Context

If the template gets large with many questions, then it is hard to find the question you are looking for.
This is adding the first usable iteration of the question filter, which filters the question list by question type or question text.

## Fixes

https://jira.esss.lu.se/browse/SWAP-1728

## Changes

Templating system editor

## Depends on

N/A


martin-trajanovski (approved):
Look great!

PR #525, opened 2021-08-17
113 contrast too low reopen

TeraMike:
## Description

I have fixed a number of other contrast issues, most relating to text that was not sufficiently dark to be properly visible.

## Motivation and Context / Fixes

The WAVE tool identified a number of problems, and I personally identified a couple that the tool did not pick up but I was only able to read by highlighting text. Now all of this text is more readable.
Issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/113

## How Has This Been Tested

Neither I or the WAVE tool identify any further contrast issues in pages or sub-menus/pages.

## Changes

Screenshots below (before above, after underneath):
Proposal workflow text darkening before and after  
Template question text darkening before and after  
SEP Members label darkening before and after  

Template metadata and topic editor before and after


TeraMike:
Waiting on this because I also need to edit the sign-in page, which uses the previous issue-causing red/pink colour scheme.

TCMeldrum:
> Waiting on this because I also need to edit the sign-in page, which uses the previous issue-causing red/pink colour scheme.

Just so you know we won't be using this sign-in page since we will be using our existing sign-in page so it is currently ESS specific

TCMeldrum (commented):
Looks good! The only thing that I have a minor problem with is the template: 
image
Should the key be slightly faded maybe? It feels like clutters the card a little and makes the key stands out slightly. I think people will identify the questions by the question and not the key. If others disagree I don't think it is that big of an issue.

TeraMike:
> Should the key be slightly faded maybe? It feels like clutters the card a little and makes the key stands out slightly. I think people will identify the questions by the question and not the key. If others disagree I don't think it is that big of an issue.

That's a fair point. I'm not able to tell between the different shades of grey myself so here are some screenshots of a few different but still WAVE-acceptable greys, let me know which you think is best:

image

Any lower and they become really hard for me personally to read. I could also un-bold them to de-emphasise them compared to the text like so:  
image

Same goes for the dependency (lightening it) though I don't think that's so significant for distracting attention. Thoughts?

TCMeldrum:
I think grey 700 or maybe 800 works. I don't think we need to unbold them but am happy to be outvoted on that one.

TickleThePanda (approved):

TCMeldrum (approved):

PR #535, opened 2021-09-02
Switch to latest version of maintained Material Table module

mr-field:
Closes UserOfficeProject/stfc-user-office-project#191

## Description
The provider for the `MaterialTable` component has been switched from the abandoned [`material-table`](https://github.com/mbrn/material-table) package to its (better) maintained fork [`@material-table/core`](https://github.com/material-table-core/core). The new package requires a more recent version of Material UI, so `@material-ui/core` has also been upgraded to the latest `4.x` version.

### Changes needed
Some of the changes required for this are simply import swaps and renaming deprecated/removed symbols. However, there are also some bigger changes:

#### Row IDs
Material Table rows are now expected to have an `id` attribute, which is used to track of the state of each row. While the documentation states that the attribute is only optional, my own testing shows that without it the table is removed from the DOM and recreated on every data update. This would cause all detail panels and a few modals to disappear under some circumstances, as they are child elements of the table.

Many of the data types the we display on Material Tables already have an `id` attribute, so no change is needed for those. However, some use a different attribute name as their "id", such as proposals with `proposalPk`. In those cases, there's currently no way to tell the table to use a different attribute as the ID. I've worked around it by copying the value of the actual ID to a new `id` attribute when binding the data. There may be cleaner ways of doing this:
- Refactor types so that the always use `id` as their unique identifier (would require a large change)
- Make a PR for the upstream package to let users set any attribute as the ID

#### Callbacks for Detail Panels
I'm not too clear if this is a bug or a feature, but Detail Panels now need to be created by a "stable function" (see https://github.com/material-table-core/core/issues/334) or they'll be remounted on table updates. I've worked around this by creating the panel in a React callback, moving the functions required by the panel into the same scope to avoid unnecessary refreshes. However, this doesn't always work - see below.

### Remaining problems
All pages seem to work OK at the moment and the tests pass, but there are still some problems with Detail Panels. Some of them (such as the ones in the Calls and Instruments tables) are still remounted on every table update despite being created in a React callback. This is very puzzling, as the nearly identical one in the SEP proposal table works fine. I haven't been able to get to the bottom of this, and it may be the same issue as https://github.com/material-table-core/core/issues/334.

I don't think this affects any functionality at the moment, so in the short term it's just a small performance issue. However, it could create problems in the future if the affected Detail Panels are changed.

## Motivation and Context
The old package was abandoned and prevented us from upgrading Material UI. It also had some accessibility issues, such as missing ARIA labels.

Changing the origin of the Material Table package allows us to move to newer Material UI versions, and gives us a chance to get fixes and improvements for Material Table in the future. It also takes care of the accessibility problems mentioned above.

## How Has This Been Tested
With the e2e tests and manual comparison of the tables before and after the change.


jekabs-karklins (approved):
Looks pretty good, besides the things already mentioned in the comments.
👍🏻

TCMeldrum (approved):
Looks good don't know what to do about the detail panel unmounting, the issue you linked looks relatively recently so hopefully, they might fix it soon?

PR #541, opened 2021-09-06
#173 Display name instead of ID in toolbar

TeraMike:
## Description

Previously `{user.email}` was being displayed, which was actually just the id in string format. This is now `{user.firstname} {user.lastname}`.

Before and after:   
image

## Motivation and Context

User experience -- UI is friendlier with a name there :)

## How Has This Been Tested

I've checked everything else which is using the UserContext to make sure it all still works and found no issues. Currently most of the use is just to access the user id so there are no issues with adding some new info to other attributes.

## Fixes

Fixes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/173

## Depends on

PR for the info being added in the backend: https://github.com/UserOfficeProject/user-office-backend/pull/400


TCMeldrum (approved):

TeraMike:
Not sure what happened with the name on this previous commit, it was intended to be "UserOfficeProject/stfc-user-office-project#173 Display preferred name if it exists"

ACLay (approved):

bolmsten (approved):

TeraMike:
Following discussion, we can just keep it as the email to avoid potentially breaking all the checks. This retains some of the friendliness (users log in with their email and it's not an arbitrary string of numbers) while not actually requiring any frontend changes at all, since `{user.email}` is what was originally displayed.

Here's what that looks like: 
image

PR #546, opened 2021-09-08
Typescript warnings about any type fixed

martin-trajanovski:
## Description

There were few typescript warnings about `any` type and this PR aims to fix that.


jekabs-karklins (approved):
👍🏻

PR #565, opened 2021-09-22
Added external auth to use a query string

TCMeldrum:
Closes UserOfficeProject/stfc-user-office-project#198

## Description

When auth redirects it encodes the sessionID as a query string, not in the URL. I have added the functionality to extract the query string if it exists with external-auth. We should standardise this to use one method but currently, we using both methods. 

One thing I am not entirely pleased with is that since the parameter in external auth is required and app.ts only uses the external auth route if a sessionID is in the URL I am currently passing a "dummy" string to satisfy these constraints. So the redirect URL from auth would be:
https://devproposal.facilities.rl.ac.uk/external-auth/sessionId?sessionid=<token>

We could change this but it would require more work.

BisAppSetting Branch: stfc-user-office-project-198-STFC-auth-redirect
Docker orchestration PR:

### DIFF
--- a/src/components/user/ExternalAuth.tsx
+++ b/src/components/user/ExternalAuth.tsx
@@ -1,5 +1,7 @@
 import PropTypes from 'prop-types';
+import queryString from 'query-string';
 import React, { useContext, useEffect, useRef } from 'react';
+import { useLocation } from 'react-router-dom';
 
 import { SettingsContext } from 'context/SettingsContextProvider';
 import { UserContext } from 'context/UserContextProvider';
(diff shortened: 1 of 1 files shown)


TickleThePanda:
> One thing I am not entirely pleased with is that since the parameter in external auth is required and app.ts only uses the external auth route if a sessionID is in the URL I am currently passing a "dummy" string to satisfy these constraints. 

On the router, could you have a route for both `/external-auth/` and `/external-auth/:session-id/` and point them both towards this component?

I believe with `Switch`, it will stop looking at other routes when it finds the first match so the order would need to be:
```
<Route path="/external-auth/:sessionId" component={ExternalAuth} />
<Route path="/external-auth/" component={ExternalAuth} />
```

or 

```
<Route exact path="/external-auth/" component={ExternalAuth} />
<Route path="/external-auth/:sessionId" component={ExternalAuth} />
```

bolmsten (approved):

TickleThePanda (approved):

PR #579, opened 2021-10-07
fix: Updating style for participator selection

jekabs-karklins:
## Description

This PR fixes some styling inconsistencies, as well as moves the  add user button down, under the table, because before it could be easily confused for the icon. 

image

## Fixes

https://jira.esss.lu.se/browse/SWAP-1867

## Depends on

N/A


martin-trajanovski (approved):
💯

PR #583, opened 2021-10-08
Last missing form label

TCMeldrum:
Closes: UserOfficeProject/stfc-user-office-project#188

## Description

Added last missing form label for the select all checkbox on a material table. Removed some selection: false as this is the default value.

## Motivation and Context

Makes the material table more usable by screen readers 

## How Has This Been Tested

Used the [wave tool](https://wave.webaim.org/)

## Fixes

UserOfficeProject/stfc-user-office-project#188

## Changes

Added an aira label through the `headerSelectionProps`  where I found material tables to have the selection option to be set to true.

### DIFF
--- a/src/components/call/AssignInstrumentsToCall.tsx
+++ b/src/components/call/AssignInstrumentsToCall.tsx
@@ -81,7 +81,7 @@ const AssignInstrumentsToCall: React.FC<AssignInstrumentsToCallProps> = ({
           search: true,
           selection: true,
           headerSelectionProps: {
-            inputProps: { 'aria-label': 'Checkbox Select All Rows' },
+            inputProps: { 'aria-label': 'Select All Rows' },
           },
           debounceInterval: 400,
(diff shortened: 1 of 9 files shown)


src/components/call/AssignInstrumentsToCall.tsx
```diff
@@ -80,6 +80,9 @@ const AssignInstrumentsToCall: React.FC<AssignInstrumentsToCallProps> = ({
         options={{
           search: true,
           selection: true,
+          headerSelectionProps: {
+            inputProps: { 'aria-label': 'Checkbox Select All Rows' },
```
mr-field, line 84: I think in general it's best to describe what the input does, rather than what the input is. A screen reader should already know that the element is a checkbox, so we should just describe it as "Select All Rows"

mr-field (commented):
Thanks for that, just thought of something else

PR #601, opened 2021-10-19
Added STFC logout

TCMeldrum:
Closes: UserOfficeProject/stfc-user-office-project#219

 
## Description
Added a call to the new logout resolver on the handleLogout ~~function when the external auth feature is enabled~~. In STFC mode this logout out the token with a call to the UOWS in other modes it does nothing as suggested [here](https://github.com/UserOfficeProject/user-office-backend/pull/446#pullrequestreview-784721538)  

~~As consequence, the user context has been moved to be above the feature context. This caused some problems with the user menu since the user menu was relying on `SettingsContextProvider` and `FeatureContextProvider` rerendering to close as when these render they [unmounts most of the frontend](https://github.com/UserOfficeProject/user-office-frontend/blob/52c575ea865f91793a5ea75045485d8b9fef10a3/src/context/FeatureContextProvider.tsx#L32) and therefore close the menu. I don't know if anywhere else relies on this fact.~~ This change has been reverted.

## Changes

- Added externalTokenLogout mutation
- Added call to external logout when logging out
- ~~Moved the Contexted around so a call can be made to the backend and feature context in the handleLogout method~~

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/446


src/graphql/user/ExternalLogoutToken.graphql
```diff
@@ -0,0 +1,8 @@
+mutation ExternalLogoutToken($externalToken: String!) {
+  ExternalLogoutToken(externalToken: $externalToken) {
```
martin-trajanovski, line 2: Just a small note here because we usually use camel case for queries and mutations and here you start with capital letter which doesn't follow our style. Is there any specific reason for it or you can change it to `externalLogoutToken` instead of `ExternalLogoutToken`?

martin-trajanovski (approved):

src/graphql/user/ExternalLogoutToken.graphql
TCMeldrum: Sorry mistake on my part I will change it

mr-field (approved):

martin-trajanovski (approved):

src/context/UserContextProvider.tsx
```diff
@@ -168,10 +170,13 @@ export const UserContextProvider: React.FC = (props): JSX.Element => {
...
-        handleLogout: useCallback(
-          () => dispatch({ type: ActionType.LOGOFFUSER, payload: null }),
-          []
-        ),
+        handleLogout: useCallback(() => {
+          unauthorizedApi().logout({
+            token: localStorage.token,
+          });
+
+          dispatch({ type: ActionType.LOGOFFUSER, payload: null });
```
jekabs-karklins, line 178: Would it not make more sense to dispatch log out in a unauthorizerdApi().logout callback?

src/context/UserContextProvider.tsx
```diff
@@ -168,10 +170,13 @@ export const UserContextProvider: React.FC = (props): JSX.Element => {
         handleLogin: (data): void =>
           dispatch({ type: ActionType.LOGINUSER, payload: data }),
         // Using useCallback here as these are used in useDataAPI dependency array
-        handleLogout: useCallback(
-          () => dispatch({ type: ActionType.LOGOFFUSER, payload: null }),
-          []
-        ),
+        handleLogout: useCallback(() => {
```
jekabs-karklins, line 173: Are we sure we need to use useCallback hook here?

jekabs-karklins (approved):

src/context/UserContextProvider.tsx
TCMeldrum: The `useCallback` hook was already there before I made the changes, trying without it seems to work fine I can update it to not use the hook.

src/context/UserContextProvider.tsx
TCMeldrum: Not sure maybe it should stay here to conform to the pattern of the `handleLogin`, `handleRole`, and `handleNewToken` functions

jekabs-karklins: Ok. I guess maybe it is even better user experience not to wait.

PR #604, opened 2021-10-21
Added Email invitation flag

TCMeldrum:
Ref: UserOfficeProject/stfc-user-office-project#238

## Description

- Email icon hidden behind Email invitation flag
- Made search box's function on people tables clearer
- Increased default page size up to 10 people

New vs Old
image
 
## Motivation and Context

For our release, we wanted to hide the email icon on the people tables as a priority but we are leaving some of the other user officer's features enabled. Also, the search box on the peoplesTables purpose was unclear so that has been updated as well as increasing the default number of people from 5 to 10.

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/447


mr-field (commented):
Just a small point of style, otherwise it looks good.

PR #642, opened 2021-11-12
Stfc users staying logged in to user office

TCMeldrum:
Closes: https://github.com/UserOfficeProject/stfc-user-office-project/issues/225

## Description

This makes it so that when a user has been logged out of the user office they will also be logged out of the proposals system. In doing this I have refactored a couple of little bits of code. `UseDataApi` now catches any errors with the `EXTERNAL_TOKEN_INVALID` message and logs out the user, while `useDataApiWithFeedback` now catches any rejection with `EXTERNAL_TOKEN_INVALID` and logs out the user. All mutations to now use `useDataApiWithFeedback` so that the rejection can be properly caught and dealt with. This also standardizes the code as only a few mutations were being called with `useDataApi` and catching the rejections to put them in the snack bar. With all cases where an `EXTERNAL_TOKEN_INVALID` error could be raised, it is caught.

## Motivation and Context

Users would stay logged in to the proposals system but would be logged out of the user office

## Fixes

https://github.com/UserOfficeProject/stfc-user-office-project/issues/225

## Changes

- Added catching EXTERNAL_TOKEN_INVALID errors 
- Deleted unused files: src/hooks/proposal/useSubmitProposal.ts and src/hooks/proposal/useCreateProposal.ts I could not find any other reference to these hooks and we seem to be submitting and creating proposals without using these hooks.
- Changes a few variable that was not following naming convention e.g. sendRequest-> api and EXTERNAL_AUTH -> externalAuth. 

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/476


src/utils/useDataApiWithFeedback.ts
```diff
@@ -26,7 +28,12 @@ function useDataApiWithFeedback() {
 
             if (result && isMutationResult(result)) {
               if (result.rejection) {
-                const { reason } = result.rejection as Rejection;
+                let { reason } = result.rejection as Rejection;
+                if (reason === 'EXTERNAL_TOKEN_INVALID') {
+                  handleLogout();
+                  reason =
+                    'You have been logged out redirecting to external homepage';
```
simonfernandes, line 35: I wonder if this might confuse some users. Would it be clearer to say that their session has expired and they need to login again?

TCMeldrum: Cool, I can update it to be clearer is `Your session has expired, you will need to log in again through the external homepage.` Although this message, depending on how fast the redirect is, might not be displayed for very long.

simonfernandes: That seems good. If it doesn't stay up that long and it looks like too much to read in time, we could probably cut the `through the external homepage` part

simonfernandes (approved):

TickleThePanda (approved):

PR #647, opened 2021-11-16
Improve performance of the instrument scientist page

TCMeldrum:
part of: https://github.com/UserOfficeProject/stfc-user-office-project/issues/254

## Description

We found that the instrument scientists were having trouble loading proposals so we wanted to optimize it. ~~We have made it fetch only proposals on the current page, not all proposals this cuts down on the time significantly. Also, cut down to just the essential data rather than all the data.~~

This cut the request down from over a minute to fetch proposals to about 7s-13s. The extra time comes from the fact that a lot of the information requested comes from different tables but am having trouble tracing the graphQL request. 

I have updated this PR to again fetch all the proposals it can but now the data comes from the proposal view so that data only comes from one source and not from multiple tables within the database. This seems to cut the query times to about 2s for now. I have removed fetching only the proposal for the current page as this was adding a 1.5-second load between pages as well as issues with sorting and searching. Some of the changes or very similar ones which they can be modeled on should be made as part of this issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/285

## Motivation and Context

https://github.com/UserOfficeProject/stfc-user-office-project/issues/254

## Depends On
https://github.com/UserOfficeProject/user-office-backend/pull/504


simonfernandes (approved):

TickleThePanda:
I think we do need to consider only loading the visible proposals. In a few months we're going to be importing at least `10 years * 2 rounds * 600 proposals` for ISIS - a much larger magnitude over what's currently in here.

Maybe it's worth trying to create that many locally via the API and seeing what the impact of the loading is - at least to get an idea of order of magnitude of loading?

Then, I think there's two options here:
 - We go ahead with this currently and then rework it in the future for ISIS.
 - We continue with improving the performance here.

There's the trade off between those options between time available now and the context that you currently have.

TCMeldrum:
I don't really mind I think this issue: https://github.com/UserOfficeProject/stfc-user-office-project/issues/285 (sorry failed to link this in the PR the first time) will be making similar changes to the ones you mentioned so I don't really want to duplicate the work by implementing a solution here as well. I am happy to wait until those changes have been made then make the changes here as well. 
I would quite like to get these changes in soon as the instrument scientist page is currently unusable.

src/components/proposal/ProposalTableInstrumentScientist.tsx
```diff
@@ -139,55 +135,22 @@ const ProposalTableInstrumentScientist: React.FC = () => {
...
-    {
-      title: 'Average Score',
-      field: 'average',
-      hidden: true,
-      render: (rowData: Proposal): number =>
-        average(getGrades(rowData.reviews)),
-      customSort: (a: Proposal, b: Proposal) =>
-        (average(getGrades(a.reviews)) || 0) -
-        (average(getGrades(b.reviews)) || 0),
-    },
```
mr-field, line 186: Looks like these columns are gone with no replacement. Are we sure we want to do that? Are they a requirement for the ESS?

src/hooks/proposal/useProposalsCoreData.ts
```diff
@@ -22,46 +25,93 @@ export function useProposalsCoreData(filter: ProposalsFilter) {
...
+            questionaryIds,
+            questionFilter: questionFilter && {
+              ...questionFilter,
+              value:
+                JSON.stringify({ value: questionFilter?.value }) ?? undefined,
+            },
+            text,
+          },
+          // offset,
+          // first,
```
mr-field, line 45: Are these meant to to be commented out? If so, can they be removed?

src/components/proposal/ProposalTableInstrumentScientist.tsx
TCMeldrum: I believe I got permission to remove these from Fredrick on the slack: https://ess-eric.slack.com/archives/C019J2U53U2/p1637655078002200?thread_ts=1637597466.002100&cid=C019J2U53U2

src/hooks/proposal/useProposalsCoreData.ts
```diff
@@ -22,46 +25,91 @@ export function useProposalsCoreData(filter: ProposalsFilter) {
...
+                JSON.stringify({ value: questionFilter?.value }) ?? undefined,
+            },
+            text,
+          },
+        })
+        .then((data) => {
+          if (unmounted) {
+            return;
+          }
+          console.log(data);
```
mr-field, line 49: I don't think we should log all raw proposal data to the console

TCMeldrum: Sorry missed that

mr-field (approved):

TCMeldrum:
@UserOfficeProject/swap Just want to check this change is OK with you before I merge

martin-trajanovski (approved):

PR #652, opened 2021-11-17
Make Proposal submission dialog text configurable

IVlaD17:
## Description
Through the changes in this PR, the text that shows up in the dialog that appears before submitting a proposal can be configured for each call. In case no text is provided, it displays a default text.

## Motivation and Context
Closes UserOfficeProject/stfc-user-office-project#261

## How Has This Been Tested
Manual front-end testing.

## Depends on
UserOfficeProject/user-office-backend#485


src/components/proposal/ProposalSummary.tsx
```diff
@@ -105,6 +105,9 @@ function ProposalReview({ confirm }: ProposalSummaryProps) {
           }
         }
       }
+      if (call?.submissionMessage != null) {
+        setSubmitButtonMessage(call.submissionMessage);
```
simonfernandes, line 109: Instead of overwriting the existing message (which dynamically says whether the proposal is editable after submission), I think we instead want to append a paragraph with this new text.

simonfernandes (changes requested):

src/components/proposal/ProposalSummary.tsx
IVlaD17: Done

jekabs-karklins (approved):

src/components/proposal/ProposalSummary.tsx
```diff
@@ -105,6 +105,11 @@ function ProposalReview({ confirm }: ProposalSummaryProps) {
           }
         }
       }
+      if (call?.submissionMessage != null) {
+        const defaultMessage =
+          'I am aware that no further edits can be done after proposal submission. \n';
```
simonfernandes, line 110: We wouldn't want the text to always say that no further edits can be done, because that won't always be the case.

I think in thise case you could rename the `checkUpcomingEditableStatus()` function to be something like `updateSubmitButtonMessage()` and have something like:

```javascript
if (proposal.status != null && hasUpcomingEditableStatus) {
  setSubmitButtonMessage(
    'Submit proposal? The proposal can be edited after submission.'.concat(
      call?.submissionMessage ? '\n' + call.submissionMessage : ''
    )
  );
} else {
  setSubmitButtonMessage(
    'I am aware that no further edits can be done after proposal submission.'.concat(
      call?.submissionMessage ? '\n' + call.submissionMessage : ''
    )
  );
}
```

simonfernandes (changes requested):

src/components/proposal/ProposalSummary.tsx
bolmsten: Can we not always just show the submission message without any prefix?

IVlaD17: My apologies, I'm not sure I follow, could you please explain?

TickleThePanda: If I've understood what Fredrik is saying - I think I prefer there being a default message with an additional message below. It means that user officers don't need to remember what the behaviour should be here. It depends on what you were thinking would be useful to add on your side here.

IVlaD17: I am sorry, I am still not following. Can you provide an example, please?

bolmsten: Okay, if you want a default one it is fine. Could we change from "I am aware that no further edits can be done after proposal submission." -> "I am aware that no further edits can be made after proposal submission."

bolmsten (approved):

simonfernandes (approved):

PR #659, opened 2021-11-19
Swap 2030 implement questionary export

jekabs-karklins:


cypress/integration/templates.ts
```diff
@@ -1543,4 +1545,135 @@ context('Template tests', () => {
         .should('have.value', 'Fig_test');
     });
   });
+
+  it('should export template in compatible format', () => {
+    const expectedExport = {
```
martin-trajanovski, line 1550: Maybe it is a better to store this expected result in a separate file under fixtures folder and just read it from there.

src/components/proposal/ProposalTableInstrumentScientist.tsx
```diff
@@ -107,7 +107,7 @@ const ProposalTableInstrumentScientist: React.FC = () => {
           </IconButton>
         </Tooltip>
 
-        <Tooltip title="Download proposal as pdf">
+        <Tooltip title="Download proposal as PFD">
```
martin-trajanovski, line 110: Just a small typo here should be `PDF`

martin-trajanovski (approved):
Looks good otherwise. 👍

PR #665, opened 2021-11-23
fix: clear arbitrary number confusion

vyshnavi1605:
## Description

Closes: UserOfficeProject/stfc-user-office-project#260
## Motivation and Context

Users may be confused by the arbitrary number provided before a proposal is submitted. When submitting a proposal, if the call has a reference number set, then the arbitrary number provided should address that fact that it is a pre-submission id. This change has been made in the dashboard and when submitting a proposal. 
## How Has This Been Tested

Tested the changes by logging into site and making 2 calls, one with a reference number and one without. Tested what the arbitrary number shows as in each call. 

image
Call with reference number set. 

image
Call without reference number set

image
Dashboard view. If reference number is set and proposal is not yet submitted, an  `*` shows beside it and is explained what it means at the bottom of the table. For the default call provided there is no `*`.
## Fixes

Fixes: UserOfficeProject/stfc-user-office-project#260
## Changes

Files changed: 
- `ProposalContainer.tsx` - added checks to see if reference number was set and rendered `info` based on that 
- `ProposalCreate.tsx` - getting call information 
- `ProposalTableUser.tsx` - added checks to see if reference number was set and rendered information based on that
- `ProposalQuestionaryReview.tsx` - added checks to see if reference number was set and rendered information based on that
- `sdk.ts` - regenerated sdk after modifying graphql files
- `graphql files` - added `referenceNumberFormat` under Call, in order to check if a reference number has been set  


src/components/review/ProposalQuestionaryReview.tsx
```diff
@@ -21,8 +21,18 @@ export default function ProposalQuestionaryReview(
 
   const users = data.users || [];
 
+  const hasReferenceNumberFormat = data.call?.referenceNumberFormat
```
TickleThePanda, line 24: Could be simplified to 
```
const hasReferenceNumberFormat = data.call?.referenceNumberFormat
```

src/components/proposal/ProposalContainer.tsx
```diff
@@ -150,17 +151,37 @@ export default function ProposalContainer(props: {
...
+
+  let info: JSX.Element | string = 'DRAFT';
+
+  if (proposalId) {
+    if (!isSubmitted && hasReferenceNumberFormat) {
+      info = (
+        <Typography>
+          {proposalId}
+          <br />
+          <small>Pre-submission-reference-number</small>
```
TickleThePanda, line 170: Let's say "Pre-submission reference" here.

src/components/proposal/ProposalTableUser.tsx
```diff
@@ -71,12 +81,15 @@ const ProposalTableUser: React.FC = () => {
...
-      isLoading={loading}
-    />
+    <div>
+      <ProposalTable
+        title="My proposals"
+        search={false}
+        searchQuery={sendUserProposalRequest}
+        isLoading={loading}
+      />
+      <br />* Pre-submission reference number, ID will change upon submission
```
TickleThePanda, line 91: "Pre-submission reference. Reference will change upon submission."

TickleThePanda (changes requested):

src/components/proposal/ProposalContainer.tsx
```diff
@@ -150,17 +151,37 @@ export default function ProposalContainer(props: {
     }
   }, [previousInitialProposal, props.proposal, dispatch]);
 
+  const hasReferenceNumberFormat = state.proposal.call?.referenceNumberFormat
+    ? true
+    : false;
+
+  const isSubmitted = state.proposal.submitted;
+
+  const proposalId = state.proposal.proposalId;
```
martin-trajanovski, line 160: Here also something like: `const { isSubmitted, proposalId } = state.proposal;` will clean up a bit so instead of two lines you can initialize both variables at once.

src/components/proposal/ProposalContainer.tsx
```diff
@@ -150,17 +151,37 @@ export default function ProposalContainer(props: {
...
+
+  let info: JSX.Element | string = 'DRAFT';
```
martin-trajanovski, line 162: My suggestion is also to simplify here a bit to something like: `let info: JSX.Element | string = proposalId || 'DRAFT';` and after that continue with: 
`
if (!isSubmitted && hasReferenceNumberFormat) {
      info = (
        <Typography>
          {proposalId}
          <br />
          <small>Pre-submission reference</small>
        </Typography>
      );
    }
`
I think it will do the same but without nesting if blocks. What you have here is also correct but it's just my suggestion to make it more readable.

src/components/proposal/ProposalTableUser.tsx
```diff
@@ -53,13 +55,21 @@ const ProposalTableUser: React.FC = () => {
               );
             })
             .map((proposal) => {
+              const hasReferenceNumberFormat = proposal.call
+                ?.referenceNumberFormat
```
martin-trajanovski, line 59: Here also using `!!roposal.call?.referenceNumberFormat` will make it a bit cleaner.

src/components/review/ProposalQuestionaryReview.tsx
```diff
@@ -21,8 +21,16 @@ export default function ProposalQuestionaryReview(
 
   const users = data.users || [];
 
+  const hasReferenceNumberFormat = data.call?.referenceNumberFormat;
```
martin-trajanovski, line 24: Also here maybe it's better to use `!!data.call?.referenceNumberFormat` to get a boolean value.

martin-trajanovski (changes requested):
Just a few small changes needed but otherwise it looks great!

martin-trajanovski (approved):

TickleThePanda (approved):

PR #698, opened 2021-12-16
Added check for reference text 

bolmsten:
## Description

This PR will add a check in the proposal table if to show the reference text based on if a proposal has a initial reference number

## Motivation and Context

Not all facilities change reference numbers after proposal submission


martin-trajanovski:
@bolmsten when you run the `generate:local` locally please make sure that your local .env file has the `SCHEMA_URL=http://localhost:4000/graphql` set to 4000 that points to the local backend not to the gateway and run `npm install` or maybe `npm ci` to have the latest packages for schema generation. Because I see some changes when I run it on my maschine.

src/components/proposal/ProposalTableUser.tsx
```diff
@@ -86,7 +86,6 @@ const ProposalTableUser: React.FC = () => {
         searchQuery={sendUserProposalRequest}
         isLoading={loading}
       />
-      <br />* Pre-submission reference. Reference will change upon submission.
```
martin-trajanovski, line 89: Should it be completely removed for users table or what was the point adding it there maybe STFC needed it.

martin-trajanovski (approved):
Just small comment otherwise it looks good.

src/components/proposal/ProposalTable.tsx
```diff
@@ -82,6 +82,14 @@ const ProposalTable = ({
     return <Redirect push to={`/ProposalEdit/${editProposalPk}`} />;
   }
 
+  const showReferenceText = (
+    proposalData: PartialProposalsDataType[]
+  ): boolean => {
+    return proposalData.some((proposal) => {
+      return proposal.call?.referenceNumberFormat && !proposal.submitted;
```
jekabs-karklins, line 89: I wonder about `&& !proposal.submitted;` part. Isn't STFC proposal editable after submission so the message should be shown until proposal is accepted and reference is updated ?

jekabs-karklins (approved):

PR #700, opened 2021-12-20
feat(shipment): fixing dialogue jerkiness when loading bar appears

jekabs-karklins:
## Description

This PR fixes the issue where when  loader appears, dialogue scrollbar shows up and hides as loader is rotating


martin-trajanovski (approved):
Looks good 👀 👍

PR #706, opened 2022-01-03
SWAP-2109: Fix organisation name when changing PI

bolmsten:
## Description

Before when changing the PI of a proposal the organisation number was shown for the old PI in the co-proposer field instead of the name.

## Motivation and Context

Reported bug

## How Has This Been Tested

locally

## Fixes

swap-2019


src/hooks/user/useUserData.ts
```diff
@@ -40,19 +44,23 @@ export function useUserWithReviewsData(filters?: {
...
+export function useBasicUserData(id?: number) {
   const api = useDataApi();
-
-  const loadBasicUserData = useCallback(
-    async (id: number) => {
-      return api()
+  const [loading, setLoading] = useState(true);
+  const [userData, setUserData] = useState<BasicUserDetails | null>(null);
+  useEffect(() => {
+    setLoading(true);
```
martin-trajanovski, line 52: Maybe it's better to set the loading if there is id. So just move inside the if statement because if id is null (which shouldn't happen but just for preventing infinite loading state) the loading will continue.

martin-trajanovski (approved):
Just a small comment otherwise it looks good.

jekabs-karklins (approved):
I like @martin-trajanovski comment. Otherwise LGTM

PR #716, opened 2022-01-09
refactor(questionary): Refactor questionary events

jekabs-karklins:
## Description

This PR reduces duplicated code in QuestionaryContainers.
There are many changes but since changes affects all 8 questionaries the actual logic that changes is not great in this PR
Questionaries:
- Proposal
- Proposal ESI
- Sample
- Sample ESI
- Feedback
- Generic templates
- Shipment
- Visit 

## Motivation and Context

In order to make it easier to add new Questionary types and maintain the old ones the code the duplicated code is reused.

## How Has This Been Tested

- [x] E2E

## Changes

- Reusing the same event for all questionaries
- Getting rid of duplicated reducers and use only one for all Questionaries

## Depends on

 N/A


martin-trajanovski (approved):
This is really good refactor and cleanup. 💯

PR #721, opened 2022-01-11
fix: validation for spinner arrows in firefox

ACLay:
## Description

Prevents the "This is a required field" validation error appearing on required number inputs in firefox when they were set using only the up/down spinner arrows.

## How Has This Been Tested

Manual testing in Chrome and Firefox on Windows 10 by clearing a required number input, attempting to save, and then setting the number via the spinner arrows. This would previously trigger the bug, but it no longer does so.

I've also tried typing numbers in the middle of the text field to ensure that the focus() call doesn't force the caret to the end or start of the textbox mid typing.

## Fixes

UserOfficeProject/stfc-user-office-project#281

## Changes

The bug was caused by the input's onBlur handler not being called, as this would normally update the component's state once it's no longer focused after being edited.
However, long standing issue in firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1012818 means that unlike chromium based browsers, firefox doesn't grant focus when spinner arrows are used, so it has no focus to lose and the blur event can't be called. To work around this we can simply force focus on the input whenever the arrows trigger a change event.

### DIFF
--- a/src/components/questionary/questionaryComponents/NumberInput/QuestionaryComponentNumberInput.tsx
+++ b/src/components/questionary/questionaryComponents/NumberInput/QuestionaryComponentNumberInput.tsx
@@ -126,6 +126,12 @@ export function QuestionaryComponentNumber(props: BasicComponentProps) {
               setStateValue(newValue);
               if (isEventFromAutoComplete(event)) {
                 onComplete(newValue);
+              } else {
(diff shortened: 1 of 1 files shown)


ACLay:
Recreating as #722 The branch name contains a `#` which is invalid in a docker tag, causing the final github action check to fail

PR #723, opened 2022-01-11
Pin Faker version to 5.5.3

mr-field:
Part of UserOfficeProject/stfc-user-office-project#312

## Description
This prevents us from picking up any future updates, now that the package has been compromised.

NPM packages are immutable once uploaded, so as long as the faker package isn't completely deleted we should be safe.

Note: this PR also reverts the package-lock to version 1. As long as we're targeting Node 12, we should keep it at version 1 as version 2 isn't fully supported.


jekabs-karklins (approved):

package.json
```diff
@@ -117,7 +117,7 @@
     "eslint-plugin-react": "^7.22.0",
     "eslint-plugin-react-hooks": "^4.2.0",
     "eslint-plugin-unused-imports": "^1.1.0",
-    "faker": "^5.3.1",
+    "faker": "5.5.3",
```
martin-trajanovski, line 120: @jekabs-karklins maybe it is worth putting a comment somewhere about this just in case we don't forget in the future.

martin-trajanovski (approved):

package.json
mr-field: Looks like there's no way to add comments here (JSON doesn't have a way of adding comments, and there doesn't seem to be a special syntax for `package.json` files). Do you have any suggestions where a comment could go? The [issue](https://github.com/UserOfficeProject/stfc-user-office-project/issues/312) will remain in Github after being closed, but I know people may not notice it.

martin-trajanovski: Yes I see. Maybe it is fine to leave it as it is for now. If the version is fixed usually someone should check the reason before upgrading.

PR #727, opened 2022-01-14
fix(shipments): Add ability to declare multiple shipments per experiment time

jekabs-karklins:
## Description

This PR aims to address the new/revised requirements for how shipments should be declared.

## Motivation and Context
Now it is possible to add multiple shipments per experiment time.
To keep things less tangled, shipment is now declared on a separate page instead of modal.

## How Has This Been Tested

- [x] Added E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2139

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/541


src/components/shipments/DeclareShipments.tsx
```diff
@@ -0,0 +1,157 @@
...
+            .then(({ shipment }) => {
+              setSelectedShipment(shipment);
+              setIsModalOpen(true);
+            })
+        }
+        onDeleteClick={onDeleteClicked}
+        onAddNewClick={hasLocalContact ? onAddClicked : undefined}
+        style={{ maxWidth: '100%' }}
+      />
+      <Divider style={{ margin: '12px 0' }} />
```
martin-trajanovski, line 128: I think it is better to use classes instead of applying styles directly. At least it is the practice that we use so far trying to keep everything in classes as much as possible.

martin-trajanovski (approved):
Just small adjustments maybe but otherwise it looks great!

PR #731, opened 2022-01-14
SWAP-2179-improve-call-modal-loading-on-open

martin-trajanovski:
## Description

Improve loading just the templates while showing the full modal on all edit or creation

## Motivation and Context

This modal loading wasn't following the same rules from the other creation/edit modals and was waiting for everything to be loaded and then show the form. It is not a good practice because the modal was in some small size while loading and then just suddenly jumping to its normal size

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2179

## Changes

https://user-images.githubusercontent.com/6333063/149537359-0cab3bee-1ae5-401a-81ee-22f822115bb0.mp4


src/hooks/call/useCallTemplates.ts
```diff
@@ -19,6 +19,7 @@ export function useActiveTemplates(
     group: groupId,
   });
 
+  // TODO: Why we are calling getTemplates twice here? This should probably be refactored a bit and use only one call.
```
jekabs-karklins, line 22: I added a PR on this that makes it  more explicit on what the code is doing. It was between this solution, or one where we fetch all templates (active and archived) templates from server and then filter out templates specified in this hook parameters.

jekabs-karklins (approved):

PR #745, opened 2022-02-01
Fix-stfc-call-creation-issue

martin-trajanovski:
## Description

- fix: Now we are using feature flag to provide correct call create/update initial values if some field is used or hidden.
- Also improved some types in the `MultistepWizard` component

## Motivation and Context

After fixing some Material-UI console warning that provided `undefined` value is out of range introduced another issue where `esiTemplateId` field is hidden under feature flag and default value (empty string '') was invalid for call creation.

## How Has This Been Tested

- manual tests


TCMeldrum (approved):
Looks like a good solution, thanks Martin!

jekabs-karklins (approved):
Converting to `undefined` could be done here here or right before making the api call.
Looks good to me

PR #747, opened 2022-02-03
Added fullWidth property to principal investigator field

nitrosx:
## Description

Principal Investigator field in Proposal creator was to narrow and would cut the text

## Motivation and Context

We would like to be able to use the full width of the form to show principal investigator info

## How Has This Been Tested

Tested locally on my machine

## Fixes

Now the Principal Investigator field on Proposal Creation form uses the full width of the form

## Changes

Added property fullWidth to relevant form component

## Depends on

No dependencies


bolmsten:
You should be able to create a branch of develop now

PR #750, opened 2022-02-04
feat: call start/end times

simonfernandes:
## Description

In the call creation dialog you can specify a time that the call will start or end. This is based on the timezone specified in a backend environment variable (e.g. 'Europe/London') and not on the timezone of the user officer creating/editing it.

To handle this, I have used Luxon time management library adapter in the Material picker. This is because `date-fns` doesn't have timezone support, and even with the `date-fns-tz` extension, it only adds some utility functions. I initially tried to get this working by implementing my own logic and overiding some `format()`, `parse()` and `label()` functions of the `date-fns` adapter, but it got very complex to implement and test. Luxon works well for this and is still being maintained unlike Moment.

Screenshots are of me based in Alaska:

## Motivation and Context

So that calls can end at a particular time instead of just the end of a day

## How Has This Been Tested

- Tested time saves correctly when creating call
- Tested time saves correctly when editing call
- Added tests for call time remaining, time field requirement

## Fixes

UserOfficeProject/stfc-user-office-project#276

## Changes

- User officer can specify a start/end time when creating/editing a call. Timezone based on env var specified.
- In the call table, start/end call time is displayed in the env var timezone.
- All other references to the call time (e.g. the deadline when a user selects a call) are based on the user's timezone
- Async job to check call end runs hourly instead of daily

## Depends on

- NodeJS backend timezone must be set using the `TZ` env var, with value of IANA timezone, e.g. `Europe/London`. 
  - For STFC: https://github.com/isisbusapps/docker-orchestration/pull/221
  - For ESS: I have only updated the e2e compose - I think this requires you to do the rest?
- Any external postgres databases must be set to an IANA timezone: `SET TIMEZONE='Europe/London';`
  - For STFC: `ops:db` in https://github.com/isisbusapps/docker-orchestration/pull/221
  - For ESS: Requires manual setting
- 


src/components/proposal/ProposalChooseCall.tsx
```diff
@@ -48,7 +38,7 @@ const ProposalChooseCall: React.FC<ProposalChooseCallProps> = ({
   };
 
   const formatDate = (date: Date) => {
-    return dateformat(new Date(date), 'dd-mmm-yyyy');
+    return dateformat(new Date(date), 'dd-mmm-yyyy h:MM TT');
```
mr-field, line 41: Is this the same format as the `dd-MMM-yyyy hh:mm a` used earlier? It would be good to use the same format as much as possible.

mr-field (commented):
How much work would it be to completely replace `date-fns` with `luxon`? It would be good to use the same date handling library everywhere, and if `date-fns` can't handle timezones it should be the one to go.

simonfernandes:
I've had a go at replacing `DateFnsUtils` with `LuxonUtils`, and I think I may have done it without too much difficulty. 

But while testing I've realised that the date comparison doesn't quite work if you want to check that a date specified is the `Exact` date of a given question (it works with the `Before` and `After` operators), which I'm struggling to fix. 

It turns out that this was already broken and isn't related to my refactoring. Do you think I should go ahead with refactoring this here and create a separate issue for the search, or keep the refactoring and issue grouped together?

image

mr-field:
> It turns out that this was already broken and isn't related to my refactoring. Do you think I should go ahead with refactoring this here and create a separate issue for the search, or keep the refactoring and issue grouped together?

Open a separate issue please, it doesn't look like it's related to the refactoring.

simonfernandes:
With https://github.com/UserOfficeProject/user-office-frontend/pull/750/commits/ba28f97fd1e4968fb044aa46d03ca78a3f66049d, I'm a bit wary of breaking something, so these are the things that I've tested below. Please let me know if there's anything else that you want me to check.

- Date with and without time works when creating proposal
- User information can be updated with a date
- Other dates in call modal save
- Searching proposal question works (apart from existing bug: https://github.com/UserOfficeProject/stfc-user-office-project/issues/332)
- Min and max date for question in template works (apart from existing bug: https://github.com/UserOfficeProject/stfc-user-office-project/issues/333)

simonfernandes:
@mr-field There's still a couple of issues with the luxon refactor that I'm struggling to fix, are you okay with me reverting that commit here and creating a separate issue?

mr-field:
> There's still a couple of issues with the luxon refactor that I'm struggling to fix, are you okay with me reverting that commit here and creating a separate issue?

That's OK with me.

PR #754, opened 2022-02-08
Update page titles#246

AspectVI:
## Description

Added Titles and Headers to every page by creating the new class Page in the Page.tsx document. This is then used to load the pages instead of 'route'. When pages are loaded this way they are able to have their title and header defined with I the Page call.

## Motivation and Context

This change makes the navigation of the site and the purpose of each page clear by uniquely naming every page based on what it performs.

Closes Update page title https://github.com/UserOfficeProject/user-office-frontend/pull/246

## How Has This Been Tested

Once changes were made a locally run version of the site was used to review the changes and see if anything broke.

## Fixes

Lack unique title and headers on pages of the proposals system

## Changes

Changes to The:
DashBoard.tsx
Page.tsx
AppToolbar.tsx


AspectVI:
Currently Unable to access Dev local site so will look in to this. (Getting error 404)

src/components/DashBoard.tsx
```diff
@@ -235,7 +241,12 @@ const Dashboard: React.FC = () => {
       </Drawer>
       <main className={classes.content}>
         <Switch>
-          <Route path="/ProposalEdit/:proposalPk" component={ProposalEdit} />
+          <Page
```
TickleThePanda, line 244: I can see there's only one route which is now a `Page`. Is that right?

AspectVI: Yeah, I noticed this and have closed this PR due to another issue anyway so will rectify this as well. All routes will be Pages.

PR #755, opened 2022-02-08
Update page title

AspectVI:
## Description

Added Titles and Headers to every page by creating the new class Page in the Page.tsx document. This is then used to load the pages instead of 'route'. When pages are loaded this way they are able to have their title and header defined with I the Page call.

## Motivation and Context

This change makes the navigation of the site and the purpose of each page clear by uniquely naming every page based on what it performs.

Closes https://github.com/UserOfficeProject/stfc-user-office-project/issues/246

## How Has This Been Tested

Once changes were made a locally run version of the site was used to review the changes and see if anything broke.

## Fixes

Lack unique title and headers on pages of the proposals system

## Changes

Changes to The:
DashBoard.tsx
Page.tsx
AppToolbar.tsx


src/components/DashBoard.tsx
```diff
@@ -235,109 +241,263 @@ const Dashboard: React.FC = () => {
...
+            setHeader={setHeader}
+            title="Samples Templates"
             path="/SampleDeclarationTemplates"
             component={SampleTemplatesPage}
           />
-          <Route path="/GenericTemplates" component={GenericTemplatesPage} />
-          <Route
+          <Page
+            setHeader={setHeader}
+            title="Generic Template"
```
TCMeldrum, line 353: There are a few inconsistencies in the naming of the pages like here "Samples Template**s**" vs "Generic Template" but other than that looks good.

AspectVI: Thanks for today Tom. This has been resolved

src/components/Page.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { useEffect } from 'react';
+import { Route, RouteProps } from 'react-router-dom';
+
+interface PageProps extends RouteProps {
+  title: string;
+  setHeader: React.Dispatch<React.SetStateAction<string>>;
+}
+
+const Page: React.FC<PageProps> = (props: PageProps) => {
```
ACLay, line 9: I'm not sure what the best name for this class would be, but I feel like it's not Page. In a web app I'd expect a Page class to model a web page, but this feels like an extension for the Route component that adds a title. `TitledRoute` maybe? `NamedRoute`?

ACLay (changes requested):

src/components/Page.tsx
AspectVI: Ok I'll rename it 'TitledRoute'

TCMeldrum (approved):
Looks good to me

ACLay (changes requested):
Just pulled the branch to have a look at how the headings appear.

I feel like we should get rid of the `| User Office / ` bit in the banner, but I realise this could be subjective.
image

There also seems to be a bit of a bug, where if you navigate to a page, then back to the root dashboard page, the old header doesn't get removed/updated.
image

It also feels a bit inconsistent that some pages have "Page" at the end of their heading but others don't. I think they'd be better without.

AspectVI:
@ACLay @TCMeldrum  Hi in relation to the 3 issues in your review;

 I feel leaving the 'User Office | ' in will help with ease of navigation but if you and tom agree it should be removed then I will remove it. 
 
 For the homepage, I will look Into it but I believe it is not created the same way as the other pages. 
 
 As for the last issue I will remove the 'page' from all headers and titles.

AspectVI:
I have looked into and made changes for 2 of Alex's Issues but the homepage not having a refreshing Title, I am having problems with. I am failing to access the Localhost version of the site and can't see my changes.

ACLay (approved):

ACLay:
It looks like your final commit switched the all the line endings in all the files you'd edited. That shouldn't be something we need to do most of the time. Trying to think what could have changed that, did you that git commit for that from windows rather than within WSL?

AspectVI:
@ACLay Is this something that needs reverting? and if so how would I do that?

PR #758, opened 2022-02-11
refactor(SEP e2e tests): Make SEP e2e tests more stable

martin-trajanovski:
## Description

Make SEP e2e tests more stable and try to avoid long screenshot name error shown.

## Motivation and Context

SEP e2e tests were failing randomly and also there was error shown for too long screenshot filenames.


martin-trajanovski:
> It looks like a lot the changes are replacing "click on this navigation element" lines with "go to this path" commands. Am I understanding that right? If that is what's going on, what was the problem with the old approach? I'd have thought that's what we'd want the tests to do, as it also tests that the required navigation sections are there and work

@ACLay Yes you are right the main change is with navigation and this navigation that you mention is just navigating to SEPs on the menu and we test that already inside `generalSeps` tests. The problem here with the random test failing was that `sep1.code` is randomly generated by faker as one word and that word sometimes can be found on a menu item or other text in the DOM. For example it generates the word `est` as a `sep1.code` and est is part of `Questions` in the menu and instead of entering specified SEP it can not find edit button there because of poor selector. I was thinking first to change every selector from `cy.contains(sep1.code).parent().find('button[title="Edit"]').click();` to `cy.get('[data-cy="SEPs-table"]').contains(sep1.code).parent().find('button[title="Edit"]').click();` but then I saw that we can just visit directly that page by id and test what is really needed in the test and not repeating and losing some time for something that is already tested.

ACLay (approved):

cypress/integration/SEP.ts
```diff
@@ -616,725 +594,673 @@ context('SEP reviews and meeting component tests', () => {
...
-        cy.contains('Meeting Components').click();
+      cy.finishedLoading();
 
-        cy.finishedLoading();
+      cy.get('[title="Show proposals"]').click();
 
-        cy.get('[title="Show proposals"]').click();
+      cy.get(
+        '[data-cy="sep-instrument-proposals-table"] tbody tr:last-child'
+      ).should('have.css', 'background-color', 'rgb(246, 104, 94)');
```
jekabs-karklins, line 871: I think it is better to add a "data-xxxxx" attribute indicating the state of the row, than tying it to the color.

jekabs-karklins (approved):

PR #759, opened 2022-02-11
feat(units): Import/export units

jekabs-karklins:
## Description

This PR aims to implement a feature to be able to transfer units from one environment to another

## Motivation and Context

Importing/exporting units makes it easier to move units around from one environment to another

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2128

## Changes

- new Units datasource
- new Units mutations/queries

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/562


src/components/common/ConflictResolver.tsx
```diff
@@ -0,0 +1,193 @@
...
+    marginRight: theme.spacing(2),
+  },
+  error: {
+    color: theme.palette.error.main,
+  },
+  resolved: {
+    color: '#ffbb00',
+  },
+  check: {
+    color: '#00ff00',
```
martin-trajanovski, line 41: Maybe here is also good to use something like success color from the palette but not sure this is just my suggestion.

src/components/common/ConflictResolver.tsx
```diff
@@ -0,0 +1,193 @@
...
+          <Check
+            className={clsx(classes.icon, classes.check)}
+            data-cy="same-icon"
+          />
+        );
+      case QuestionComparisonStatus.SAME:
+        return (
+          <DoneAllIcon
+            className={clsx(classes.icon, classes.check)}
+            data-cy="same-icon"
```
martin-trajanovski, line 102: I see that `DoneAllIcon` and the previous `Check` icon have the same `data-cy` attribute. Maybe this is fine but I think it is better to be different.

src/components/questionary/questionaryComponents/Interval/QuestionaryComponentInterval.tsx
```diff
@@ -132,6 +132,7 @@ export function QuestionaryComponentInterval(props: BasicComponentProps) {
             label="Min"
             id={`${id}-Min`}
             onChange={(event) => {
+              // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
```
martin-trajanovski, line 135: Not sure if you need this comment at all because I don't see any non-null assertion.

src/components/questionary/questionaryComponents/Interval/QuestionaryComponentInterval.tsx
```diff
@@ -160,6 +161,7 @@ export function QuestionaryComponentInterval(props: BasicComponentProps) {
             label="Max"
             id={`${id}-Max`}
             onChange={(event) => {
+              // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
```
martin-trajanovski, line 164: Same here.

src/components/questionary/questionaryComponents/NumberInput/QuestionaryComponentNumberInput.tsx
```diff
@@ -132,6 +132,7 @@ export function QuestionaryComponentNumber(props: BasicComponentProps) {
             label="Value"
             id={`${id}-value`}
             onChange={(event) => {
+              // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain
```
martin-trajanovski, line 135: Same here.

src/components/settings/unitList/CreateUnit.tsx
```diff
@@ -98,7 +98,6 @@ const CreateUnit: React.FC<CreateUnitProps> = ({ close, unit }) => {
             }))}
             InputProps={{ 'data-cy': 'unit-quantity' }}
             required
-            loading={loadingQuantities}
```
martin-trajanovski, line 101: Wasn't it better UX when you show the form and just loading on the autocomplete until quantities are loaded. Because like this the view jumps I guess from loading to form show after loading. Maybe there is a reason why you changed it but just wanted to be sure.

src/components/unit/ImportUnitsPage.tsx
```diff
@@ -0,0 +1,54 @@
...
+  });
+};
+
+export default function ImportUnitsPage() {
+  const { api } = useDataApiWithFeedback();
+  const [validationResult, setValidationResult] =
+    React.useState<UnitsImportWithValidation | null>(null);
+
+  return (
+    <Container>
```
martin-trajanovski, line 27: I see that we usually use `StyledContainer` at the pages like this but maybe here you have some reason not to use it. This is because of alignment of paddings and margins with other views. Maybe here it is already aligned like this but just wanted to be sure.

src/components/unit/ImportUnitsPage.tsx
```diff
@@ -0,0 +1,54 @@
...
+        </Typography>
+        {validationResult ? (
+          <UnitMergeReview
+            data={validationResult}
+            onBack={() => setValidationResult(null)}
+          />
+        ) : (
+          <SelectImportFile
+            onFileSelected={(json) => {
+              api()
```
martin-trajanovski, line 40: I think it is much better to keep the logic outside the html(tsx) in a separate function and just call it here. It looks much cleaner.

src/units/UnitMergeReview.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+      {props.data.unitComparisons.map((comparison) => (
+        <ConflictResolver<UnitComparison>
+          key={comparison.newUnit.id}
+          comparison={comparison}
+          onConflictResolved={onConflictResolved}
+          getStatus={(comparison) => comparison.status}
+          getItemId={(comparison) => comparison.newUnit.id}
+          getItemTitle={(comparison) => comparison.newUnit.unit}
+          getDiffInfo={({ existingUnit, newUnit }) => {
```
martin-trajanovski, line 97: Here also this returning array would be better to keep in a function and just call it here. My point is to clean up the html(tsx) part as much as possible because it looks better.

martin-trajanovski (commented):
Looks good! I have left few small review comments to check.

TickleThePanda (approved):

src/components/common/ConflictResolver.tsx
jekabs-karklins: Sounds like a good idea

src/components/settings/unitList/CreateUnit.tsx
jekabs-karklins: It's because if units fail to load you end up with form that does not work. It's better to not to show broken form to a user at all, I though.

martin-trajanovski: I still think there could be some improvement but it is not so important at the moment.

martin-trajanovski (approved):

vyshnavi1605 (approved):

PR #761, opened 2022-02-15
feat: Allow to unsubmit proposal

jekabs-karklins:
## Description

This PR aims to address the issue that the proposal needs to be reopened after the call has ended

## Motivation and Context

NOTE: If you want to have the possibility to reopen proposal after submission you need to have EDITABLE_SUBMITTED state somewhere in your workflow.

If user wants to have the proposal reopened for editing then User officer needs to set the status of the proposal to `EDITABLE_SUBMITTED`. Then user can edit the proposal even if proposal is submitted and the call has ended. When user is done with the changes, user officer needs to set back the appropriate status for the proposal.

NOTE: This is somewhat temporary solution which works fine but could become tedious if proposal count is large. The more streamlined solution will be added once the advanced workflow configuration UI is implemented.

## How Has This Been Tested

Added tests to cover the changes

## Fixes

https://jira.esss.lu.se/browse/SWAP-2253

## Depends on
https://github.com/UserOfficeProject/user-office-backend/pull/566


martin-trajanovski (approved):
Looks good to me.

PR #763, opened 2022-02-17
fix: Hide incompatible calls

jekabs-karklins:
## Description

This PR aims to address the issue where we can not clone proposal on a call whose template is different. This fix makes sure the incompatible calls are not visible to the user

## How Has This Been Tested

- [x] Manual tests
- [x] E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2301

## Changes

Call selector

## Depends on

N/A


src/components/proposal/ProposalTable.tsx
```diff
@@ -179,11 +182,13 @@ const ProposalTable = ({
           {
             icon: FileCopy,
             tooltip: 'Clone proposal',
-            onClick: (event, rowData) => {
-              setProposalToCloneId(
-                (rowData as PartialProposalsDataType).primaryKey
-              );
-              setOpenCallSelection(true);
+            onClick: (_event, rowData) => {
+              api()
```
martin-trajanovski, line 186: I think it is much better to extract this onClick logic/api call as a separate function outside of the `tsx` just for readability.

TickleThePanda (approved):

src/components/proposal/ProposalTable.tsx
jekabs-karklins: The ProposalsTable should be refactored and all actions should be moved out. Did not want to make it part of this PR.

martin-trajanovski (approved):

PR #768, opened 2022-02-21
Optimise  user officer proposal table

keiranjprice101:
## Description

This PR changes the user officer proposal table to not load all proposals at once, but instead load and unload 200 at a time. 
It's also my first ever react/typescript PR so be extra vigilant 😉 

## Motivation and Context
This change is necessary to allow historic STFC proposals to be loaded in without bringing the user office table to a standstill. With an estimate of at least 15,000 proposals to be added (700 x 2 rounds x 10 years). When as little as 5000 Dummy proposals were created the user officer table became unusable on dev.

## How Has This Been Tested

Manually tested and all e2e tests are passing/updated.

## Fixes

Closes UserOfficeProject/user-office-project-issue-tracker#285

## Changes
Some of the changes feel more on the hacky side, this is due to the [material-table-core framework](https://github.com/material-table-core/core). While it does claim to support remote data, a lot of features do not work well or at all.  (The library is a community fork of [material-table](https://github.com/mbrn/material-table), and the fork has[ no documentation](https://material-table-core.com/docs/api/props/mt_prop_data) and the documentation for the original library is poor and possibly outdated in areas).

Because we use custom react hooks to access the API, these cannot be given to the table component through the data prop. To attempt to make that work a lot of code would need to be duplicated to create an async function outside of the hooks to use this prop. Even if this was done the sorting and searching does not seem to be possible to implement in the same way (this is even true in the material table documentation, where searching and filtering does not work properly in the remote example)

A lot of the above issues and others could be mitigated in the future by considering a different table library.

The only feature that seems to be lost is maintaining a selection if the table data needs to be reloaded. There does not seem to be a trivial way to implement this as the table data gets changed each time it reloads.

The sorting and search has been moved into the backend.

I will add more relevant and specific inline comments below.

## Depends on

UserOfficeProject/user-office-backend#568


cypress/integration/proposal_administration.ts
```diff
@@ -475,7 +475,9 @@ context('Proposal administration tests', () => {
       cy.contains('Title')
         .parent()
         .find('[data-testid="mtableheader-sortlabel"]')
-        .dblclick();
+        .click();
```
keiranjprice101, line 478: Double click to quickly reverse the order cannot work on the user officer table, as the sort is now performed in the backend, meaning the table reloads after the first click.

cypress/integration/proposal_administration.ts
```diff
@@ -531,13 +533,13 @@ context('Proposal administration tests', () => {
 
       cy.finishedLoading();
 
-      cy.get('table tbody tr').eq(0).contains(proposalFixedName);
-      cy.contains('Title').dblclick();
       cy.get('table tbody tr').eq(1).contains(proposalFixedName);
+      cy.contains('Title').click();
```
keiranjprice101, line 537: I couldn't figure out why the table is populated in a different order with this change (I'm not sure how the original results were ordered, but if someone knows I can always set a default sort) but sorting in the other order and reversing the assertions tests the same functionality.

src/components/common/SuperMaterialTable.tsx
```diff
@@ -23,13 +23,15 @@ export type UrlQueryParamsType = {
   selection: QueryParamConfig<(string | null | never)[]>;
   sortColumn: QueryParamConfig<number | null | undefined>;
   sortDirection: QueryParamConfig<string | null | undefined>;
+  sortField?: QueryParamConfig<string | null | undefined>;
```
keiranjprice101, line 26: Due to the order in which the columns and table are built and rendered, restoring a sort via the column id would result in querying the backend twice when the page initially loads. So I've added the field to the url params to prevent this while still allowing sorts to be applied from the query params.

src/hooks/proposal/useProposalsCoreData.ts
```diff
@@ -8,7 +8,14 @@ import { UserContext } from 'context/UserContextProvider';
 import { ProposalsFilter, ProposalView, UserRole } from 'generated/sdk';
 import { useDataApi } from 'hooks/common/useDataApi';
 
-export function useProposalsCoreData(filter: ProposalsFilter) {
+export function useProposalsCoreData(
+  filter: ProposalsFilter,
+  first?: number,
```
jekabs-karklins, line 13: I see in a few places you have the same set of parameters passed around. I wonder if it is not easier to define a typed object and pass that around. 

```
first?: number,
offset?: number,
sortField?: string,
sortDirection?: string,
searchText?: string
```

jekabs-karklins (approved):
Check comments, but otherwise looks good

src/hooks/proposal/useProposalsCoreData.ts
keiranjprice101: Thanks, I could use the query object directly to be more succinct 👍

TCMeldrum (approved):

PR #775, opened 2022-02-24
feat: Add start and end visit dates

jekabs-karklins:
## Description

This PR aims to add start and end visit dates to the database. This information, later on, will be shown on the experiments page

## Motivation and Context

In the Experiments Page, we will need to show visits along with their start and end dates.

## How Has This Been Tested

Added and updated E2E

## Changes

- visit questionary components

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/570


src/components/visit/VisitRegistrationReview.tsx
```diff
@@ -1,3 +1,4 @@
+import moment from 'moment';
```
martin-trajanovski, line 1: I honestly think that we should move away from using `moment` because `luxon` was introduced by STFC people as better library to work with when using dates and times and also it has better support. I see that the moment is not even part of our direct dependencies in the `package.json` but `luxon` is, so I guess there is all the functionality that we need. There are few other places that we should change to use luxon instead of moment but that could come with another PR.

src/components/questionary/questionaryComponents/VisitBasis/QuestionaryComponentVisitBasis.tsx
```diff
@@ -1,14 +1,39 @@
+import DateFnsUtils from '@date-io/date-fns';
```
martin-trajanovski, line 1: Also about `DateFnsUtils` I think it is better to use `LuxonUtils` as it is done here: https://github.com/UserOfficeProject/user-office-frontend/blob/develop/src/components/call/CallGeneralInfo.tsx#L160 and slowly but surely move away from any other library used for dates other than luxon. It is better I think to have only one library for this and use it everywhere.

martin-trajanovski (changes requested):
Looks good just few small change to standardize one date/time library usage.

src/components/visit/VisitRegistrationReview.tsx
jekabs-karklins: Yes, I think we should have abstraction util for time methods so that if we decide to switch either library or time format or have it controlled differently per site there is one central place to do it.

martin-trajanovski (approved):
Looks good now 👍

PR #782, opened 2022-02-25
refactor: make ProposalEsiDetails component reusable

jekabs-karklins:
## Description

This PR aims to make ProposalEsiDetails component reusable so that it can be reused in other places

## Motivation and Context

Experiments table will need this component

## How Has This Been Tested

- [x] E2E tests
- [x] Manual tests

## Changes

- ProposalEsiDetails
- ProposalEsiReview.tsx
- ProposalEsiDetailsButton.tsx
- ButtonWithDialog.tsx

## Depends on

N/A


src/hooks/common/ButtonWithDialog.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface ButtonWithDialogProps {
+  children: JSX.Element;
+  label: string;
+}
+
+function ButtonWithDialog({ children, label }: ButtonWithDialogProps) {
+  const [isDialogOpen, setIsDialogOpen] = React.useState(false);
+
+  return (
+    <div>
```
martin-trajanovski, line 16: You can use react fragment here `<></>` instead of div if it not really needed.

martin-trajanovski (approved):
Looks good just a small comment

PR #783, opened 2022-02-25
feat(SEPs): Add submitted checkbox inside SEP review grading for officer and SEP Chair/Secretary to be able to un-submit

martin-trajanovski:
## Description

This PR allows User officer, SEP Chair and SEP Secretary to un-submit SEP review.

## Motivation and Context

Previously no one was able to unsubmit submitted SEP review.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2316

## Changes

https://user-images.githubusercontent.com/6333063/157194093-9a3530f6-1f2e-40c0-b412-59f67b7c2e21.mp4

## Depends on

Backend PR: https://github.com/UserOfficeProject/user-office-backend/pull/571


cypress/integration/SEP.ts
```diff
@@ -14,7 +14,7 @@ const sepMembers = {
   reviewer2: initialDBData.users.user3,
 };
 
-function readWriteReview() {
+function readWriteReview({ shouldSubmit } = { shouldSubmit: false }) {
```
jekabs-karklins, line 17: could this be just like?
`readWriteReview(shouldSubmit: false)`

jekabs-karklins (approved):

cypress/integration/SEP.ts
martin-trajanovski: It could be something like: `readWriteReview(shouldSubmit = false)` but my point here is to be more readable when you call the function that is why I want to have parameter as an object and not as a variable. For example I find this (`readWriteReview({ shouldSubmit: true })`) more readable then this (`readWriteReview(true)`). In the second situation you don't really know what this true is exactly doing if you don't check the definition of the function.

PR #787, opened 2022-03-01
feat: Show experiment times

jekabs-karklins:
## Description

This PR aims to implement a feature where the user offer has an overview of what are the upcoming experiments. 
In this view, officers can see also visit information, and also proposal ESI.

This view could be extended to add more relevant information for user officers to see other information about upcoming experiments

Peek 2022-03-03 14-04

## Motivation and Context

Currently, there was no way for the user officer to see the list of upcoming experiments or visits

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-1635

## Changes

ScheduledEventDataSource new endpoints

## Depends on
https://github.com/UserOfficeProject/user-office-backend/pull/576


martin-trajanovski (changes requested):
Looks fine with some small changes needed.

cypress/integration/experiments.ts
```diff
@@ -0,0 +1,91 @@
...
+  before(() => {
+    cy.resetDB(true);
+  });
+
+  beforeEach(() => {
+    cy.viewport(1920, 1080);
+    cy.resetDB(true);
+    cy.updateProposalManagementDecision({
+      proposalPk: initialDBData.proposal.id,
+      statusId: 1,
```
martin-trajanovski, line 13: Probably we can use `ProposalEndStatus` enum here instead of just a number for better readability.

src/components/experiment/DateFilter.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+function DateFilter(props: DateFilterProps) {
+  const classes = useStyles();
+  const [presetValue, setPresetValue] = React.useState<TimeSpan | null>(null);
+  const { settings } = useContext(SettingsContext);
+  const inputDateFormat =
+    settings.get(SettingsId.DATE_FORMAT)?.settingsValue ?? 'yyyy-MM-dd';
+
+  return (
+    <FormControl className={classes.formControl}>
```
martin-trajanovski, line 76: Probably using `Stack` with some attributes here it better option instead of `FormControl` but you should test not quite sure.

src/components/experiment/DateFilter.tsx
```diff
@@ -0,0 +1,125 @@
...
+  to?: Date;
+  onChange: (from?: Date, to?: Date) => void;
+}
+
+function DateFilter(props: DateFilterProps) {
+  const classes = useStyles();
+  const [presetValue, setPresetValue] = React.useState<TimeSpan | null>(null);
+  const { settings } = useContext(SettingsContext);
+  const inputDateFormat =
+    settings.get(SettingsId.DATE_FORMAT)?.settingsValue ?? 'yyyy-MM-dd';
```
martin-trajanovski, line 73: Using the hook `useFormattedDateTime` should be better here instead of getting the format from the settings. You should pass parameter: `settingsFormatToUse` as `Settings.DATE_FORMAT`.

src/components/experiment/ExperimentFilterBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <>
+      <CallFilter
+        callId={urlQueryParams.call ?? undefined}
+        calls={calls}
+        isLoading={loadingCalls}
+        shouldShowAll={true}
+        onChange={(callId) => {
+          setUrlQueryParams({ call: callId });
```
martin-trajanovski, line 34: I think this `CallFilter` component is already setting the `call` in the url query string so you don't need to re-set it again. If you see onChange inside the component will be more clear.

src/components/experiment/ExperimentFilterBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+      to: DateParam,
+    });
+
+  const { instruments, loadingInstruments } = useInstrumentsData();
+  const { calls, loadingCalls } = useCallsData();
+
+  return (
+    <>
+      <CallFilter
+        callId={urlQueryParams.call ?? undefined}
```
martin-trajanovski, line 29: Maybe it is better option to use `withDefault(NumberParam, undefined)` when defining call inside the `useQueryParams` instead of this check here.

src/components/experiment/ExperimentFilterBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+        }}
+        data-cy="call-filter"
+      />
+
+      <InstrumentFilter
+        instrumentId={urlQueryParams.instrument ?? undefined}
+        instruments={instruments}
+        isLoading={loadingInstruments}
+        shouldShowAll={true}
+        onChange={(instrumentId) => {
```
martin-trajanovski, line 44: `InstrumentFilter` component is also setting the query parameter inside it's onChange function so this is repeating the same operation twice.

src/components/experiment/ExperimentFilterBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+          setUrlQueryParams({ instrument: instrumentId });
+        }}
+        data-cy="instrument-filter"
+      />
+
+      <DateFilter
+        from={urlQueryParams.from ?? undefined}
+        to={urlQueryParams.to ?? undefined}
+        onChange={(from, to) => {
```
martin-trajanovski, line 53: `DateFilter` component should probably follow the same path with setting the query in it's own onChange function.

src/components/experiment/ExperimentPage.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { StyledContainer, StyledPaper } from 'styles/StyledComponents';
+
+import ExperimentFilterBar from './ExperimentFilterBar';
+import ExperimentTable from './ExperimentTable';
+
+function ExperimentPage() {
+  return (
+    <StyledContainer>
+      <Grid container>
```
martin-trajanovski, line 12: Grid container and item not needed here as we removed them to follow some better standard for all pages.

src/components/experiment/ExperimentTable.tsx
```diff
@@ -0,0 +1,135 @@
...
+  },
+  {
+    title: 'Proposal',
+    field: 'proposal.title',
+  },
+  {
+    title: 'Experiment start',
+    field: 'startsAt',
+    render: (rowData: RowType) =>
+      DateTime.fromFormat(rowData.startsAt, 'yyyy-MM-dd HH:mm:ss').toFormat(
```
martin-trajanovski, line 37: Formating here should be done using the standard formats. Something like here: https://github.com/UserOfficeProject/user-office-frontend/blob/develop/src/components/call/CallsTable.tsx#L218

src/components/experiment/ExperimentTable.tsx
```diff
@@ -0,0 +1,135 @@
...
+        'dd/MM/yyyy HH:mm'
+      ),
+  },
+  {
+    title: 'Experiment end',
+    field: 'endsAt',
+    render: (rowData: RowType) =>
+      DateTime.fromFormat(rowData.endsAt, 'yyyy-MM-dd HH:mm:ss').toFormat(
```
martin-trajanovski, line 45: The same goes here

src/components/experiment/ExperimentVisitsTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+}));
+
+type RowType = NonNullable<
+  GetScheduledEventsCoreQuery['scheduledEventsCore'][0]['visit']
+>['registrations'][0];
+
+const getHumanReadableStatus = (rowData: RowType) => {
+  switch (rowData.trainingStatus) {
+    case TrainingStatus.ACTIVE:
+      return `Valid until ${DateTime.fromISO(
```
martin-trajanovski, line 33: Formatting here as well should be using `useFomrattedDateTime`.

src/components/experiment/ExperimentVisitsTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+        rowData.trainingExpiryDate
+      ).toFormat('dd-MM-yyyy HH:mm')}`;
+    case TrainingStatus.EXPIRED:
+      return `Expired ${DateTime.fromISO(rowData.trainingExpiryDate).toFormat(
+        'dd-MM-yyyy HH:mm'
```
martin-trajanovski, line 38: Here as well

src/components/experiment/ExperimentVisitsTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+      title: 'Teamleader',
+      render: (rowData: RowType) =>
+        rowData.userId === scheduledEvent.visit?.teamLead.id ? 'Yes' : 'No',
+    },
+    {
+      title: 'Starts at',
+      field: 'rowData.startsAt',
+      render: (rowData: RowType) =>
+        rowData.isRegistrationSubmitted
+          ? formatDate(rowData.startsAt)
```
martin-trajanovski, line 76: Here as well as mentioned above should follow this example: https://github.com/UserOfficeProject/user-office-frontend/blob/develop/src/components/call/CallsTable.tsx#L218

src/components/experiment/ExperimentVisitsTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+          : 'Unfinished',
+    },
+    {
+      title: 'Ends at',
+      field: 'rowData.endsAt',
+      render: (rowData: RowType) =>
+        rowData.isRegistrationSubmitted
+          ? formatDate(rowData.endsAt)
```
martin-trajanovski, line 84: here as well

src/components/proposalBooking/CreateUpdateVisit.tsx
```diff
@@ -112,7 +112,9 @@ function CreateUpdateVisit({ event, close }: CreateUpdateVisitProps) {
             <Button
               disabled={isSubmitting}
               type="submit"
-              data-cy="create-visit-button"
+              variant="contained"
```
martin-trajanovski, line 115: Probably variant and color are not needed here because we are setting those as defaults in the theme.ts

src/hooks/common/ButtonWithDialog.tsx
```diff
@@ -9,26 +11,42 @@ interface ButtonWithDialogProps {
...
 function ButtonWithDialog({ children, label }: ButtonWithDialogProps) {
   const [isDialogOpen, setIsDialogOpen] = React.useState(false);
+  const classes = useStyles();
 
   return (
     <>
-      <Button variant="text" onClick={() => setIsDialogOpen(true)}>
+      <Link
+        onClick={() => setIsDialogOpen(true)}
+        color="primary"
```
martin-trajanovski, line 31: Color primary not needed as that is the default color anyway.

martin-trajanovski (changes requested):

cypress/integration/experiments.ts
jekabs-karklins: I think only finalStatus has a datatype of ProposalEndStatus. The statusId is of type number

src/components/experiment/ExperimentFilterBar.tsx
jekabs-karklins: I think this is fine. It can be null if users want's to get all results, or undefined if he is not setting anything. If using with default option, I need to exclude the `null` from `ExperimentUrlQueryParamsType['call']` type. This also makes it look less consistent with other parameters.

src/components/experiment/DateFilter.tsx
jekabs-karklins: 🆗 
BTW we have many other places where this should be refactored
I added the task here https://jira.esss.lu.se/browse/SWAP-2504

src/components/experiment/ExperimentTable.tsx
```diff
@@ -100,6 +56,48 @@ function ExperimentTable() {
...
+    },
+    {
+      title: `Experiment start (${timezone})`,
+      field: 'startsAt',
+      render: (rowData: RowType) => toFormattedDateTime(rowData.startsAt),
+    },
+    {
+      title: `Experiment end (${timezone})`,
+      field: 'endsAt',
+      render: (rowData: RowType) => toFormattedDateTime(rowData.endsAt),
```
martin-trajanovski, line 84: This will probably result in a console warning from material-table because that is how it works for some reason. That's why I suggested keeping the columns outside the component but just maping here inside and adding the new fields that are formatted dates. Something like it is done here: https://github.com/UserOfficeProject/user-office-frontend/blob/develop/src/components/call/CallsTable.tsx#L218

src/components/experiment/ExperimentVisitsTable.tsx
```diff
@@ -73,15 +69,15 @@ function ExperimentVisitsTable({
       field: 'rowData.startsAt',
       render: (rowData: RowType) =>
         rowData.isRegistrationSubmitted
-          ? formatDate(rowData.startsAt)
+          ? toFormattedDateTime(rowData.startsAt)
```
martin-trajanovski, line 72: Here comes the same situation with the warning thing. If we use render inside the component it always gives a warning that should not re-render or whatever. The easiest fix is to keep columns outside of the component and without render option but instead something like this component: https://github.com/UserOfficeProject/user-office-frontend/blob/develop/src/components/call/CallsTable.tsx#L218

martin-trajanovski (approved):

PR #804, opened 2022-03-09
fix: improve questionary look and feel

jekabs-karklins:
## Description
This PR aims to improve the UI of questionary, 
- by adjusting the spacing
- grouping
- improving consistency
- providing users with relevant information

### Old look vs new look small UI tweaks
image
image

### Clicking proposalID copies it to clipboard
image

### Sample declaration
image
image

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2361

## Changes

Questionary components

## Depends on

N/A


src/components/common/CopyToClipboard.tsx
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  return (
+    <Tooltip
+      onClick={handleClick}
+      color="primary"
+      title="Copy to clipboard"
+      className={classes.withPointer}
+    >
+      <div className={classes.container}>{children}</div>
```
martin-trajanovski, line 38: I think in cases like this one it is much better to use something like `<Box display="inline-block">{children}</Box>` instead of using div and defining custom class just for the display property. I think we should start using the `Box` component more because it accepts many customizations.

src/components/questionary/questionaryComponents/NumberInput/QuestionaryComponentNumberInput.tsx
```diff
@@ -129,7 +127,6 @@ export function QuestionaryComponentNumber(props: BasicComponentProps) {
         </Grid>
         <Grid item xs={2} className={classes.unitField}>
           <TextField
-            label="Value"
```
martin-trajanovski, line 132: Not sure why we remove the label property here. It is a bit strange to see input without label.

src/components/questionary/questionaryComponents/QuestionnairesList.tsx
```diff
@@ -47,42 +60,50 @@ export function QuestionnairesList({
...
+                onEditClick={onEditClick}
+                onDeleteClick={onDeleteClick}
+                onCloneClick={onCloneClick}
+                key={record.id}
+              />
+            );
+          })}
+        </List>
+      )}
+      <div className={classes.bottomBar}>
```
martin-trajanovski, line 88: The same with the Box component goes here it is easier to understand it it is written like: `<Box display="flex" alignItems="center">`

src/components/questionary/questionaryComponents/SampleDeclaration/QuestionaryComponentSampleDeclaration.tsx
```diff
@@ -121,116 +125,118 @@ function QuestionaryComponentSampleDeclaration(
             });
 
         return (
-          <div>
+          <>
```
martin-trajanovski, line 128: Nice one 👍

martin-trajanovski (changes requested):
Just small fixes but otherwise it looks nice.

src/components/common/CopyToClipboard.tsx
jekabs-karklins: I have changed this component a little and not using ToolTip
image
I think too that we could use Box more as it has a lot of CSS as attributes, so we can reduce amount of makeStyles

martin-trajanovski (approved):

PR #808, opened 2022-03-10
Firefox does not set 'sameSite' by default for cookies

TCMeldrum:
Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/311

## Description

Firefox does not set the sameSite header in cookies by default causing a warning in the console. It seems like it defaults to 'lax' on other browsers so setting explicitly fixes the error.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite


keiranjprice101 (approved):

vyshnavi1605 (approved):

src/context/UserContextProvider.tsx
```diff
@@ -163,6 +163,7 @@ export const UserContextProvider: React.FC = (props): JSX.Element => {
       // don't support .domain in browsers while setting cookies
       // include the leading dot only for "real" domains
       domain: hostname.includes('.') ? `.${hostname}` : hostname,
+      sameSite: 'lax',
```
martin-trajanovski, line 166: Is this set just to cover older Firefox versions because I see that now default value is `lax` anyway. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

martin-trajanovski (approved):

src/context/UserContextProvider.tsx
TCMeldrum: Unless I am misinterpreting https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#browser_compatibility looks like firefox has does not set `sameSite` by default as it is an optional setting. Also looks like it is also disabled in safari so might still be useful anyway?

PR #810, opened 2022-03-11
fix: questions list not updated once question is deleted

jekabs-karklins:
## Description

This PR aims to address the bug where questions would disappear from the list once deleted
image
## Motivation and Context

Bugfix

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2319

## Changes

- delete handler return value was incorrect. Now returning the correct value so that SuperMateralTable can remove the element from the list
- Improved also Units table
- Added test for testing deleting questions
- Refactored question E2E to keep the style consistent

## Depends on

N/A


martin-trajanovski (approved):
Looks really nice!

PR #785, opened 2022-02-28
refactor(material-ui): Upgrade MUI to v5 and use only luxon for datetime manipulation

martin-trajanovski:
## Description

- Upgrade of `material-ui` from v4 to v5
- Upgraded `@material-table/core` to version compatible with MUI 5.
- Using only `luxon` for datetime manipulation instead of mixing couple of other libraries.
- Using some global theme settings to set default properties on fields instead of repeating on each and every form control.

## Motivation and Context

To be able to use and develop `DateTimeRangePicker` in the future it was recommended to upgrade to MUI 5 because the `material-ui-pickers` package is deprecated(https://github.com/mui-org/material-ui-pickers/issues/2157) and now pickers are part of `@mui/lab` official package.
 
## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

- https://jira.esss.lu.se/browse/SWAP-1937
- https://jira.esss.lu.se/browse/SWAP-2379

## Tests included/Docs Updated?

- [x] I have added tests to cover my changes.
- [x] All relevant doc has been updated

PR #813, opened 2022-03-14
Refactor date adapter to use luxon

TCMeldrum:
Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/335

## Description

These changes were previously added in https://github.com/UserOfficeProject/user-office-frontend/pull/750 (See the [issue](https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/335) for more details) but there were some issues with the automatic tests which I have now hopefully fixed here,

Before we were using a moment to convert the date to ISO format which seemed to cause issues when interacting with luxon so instead we are now using a inbuilt luxon function. There is also a weird bug when deleting a date caused it to be refilled with the default date. I have refactored the code to send over a string of `Invalid Date` on the onComplet function like date-fns did (Luxon sends a date object with a invalidate flag) and to explicitly return `null`. This seemed to have fixed it.

There still seems to be some weirdness with the date picker as spamming the save button can reset the date to the default/current date and save it when the date is out of the required range. But this seems to be not an issue with the refactor. 


martin-trajanovski:
@TCMeldrum this is related to https://github.com/UserOfficeProject/user-office-frontend/pull/815. I did biggest refactor with upgrading the material-ui to v5 here (https://github.com/UserOfficeProject/user-office-frontend/pull/785) and continued with time format standardization and moving completely to luxon. Probably this was a bit of a lack in communication but if you merge this one there will be many conflicts because it is already done in the PR that I mentioned.

TCMeldrum:
OK cool no worries if this is fixed in your refactor I am happy to close this PR.

martin-trajanovski:
@TCMeldrum thank you and sorry for the inconvenience.

PR #815, opened 2022-03-14
feat: Standardize date format using database setting for it all around the app. Also fix some date fields issues

martin-trajanovski:
## Description

This PR aims to standardize the date and datetime formats used in the user office software. Previously we were modifying the dates a lot with different kind of formats throughout the app.
- Remove custom `TzLessDateTime` scalar type, use `luxon` instead of moment for dates
- Now all the formats should be kept as setting in the database and use for rendering dates on the frontend in the same format.
- This also fixes some issues with date search filter when exact date is used to filter by.

## Motivation and Context

Have one standard for rendering dates and keep that at one place for easier maintenance and easier modifications based on facilities.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2345

## Depends on

Backend PR: https://github.com/UserOfficeProject/user-office-backend/pull/597


keiranjprice101 (approved):

jekabs-karklins (approved):
It is a very large refactor, but overall looks good
💯

sc19ew (approved):

PR #820, opened 2022-03-18
Query proposals by reference number

keiranjprice101:
## Description
Part of https://github.com/UserOfficeProject/user-office-backend/pull/604


martin-trajanovski (commented):
Looks fine but not sure if we are using this filter anywhere or it is just added.

keiranjprice101:
@martin-trajanovski that's correct, it will be used externally.

martin-trajanovski (approved):

PR #824, opened 2022-03-21
fix: Links to be consistent and logo to be shown

bolmsten:
## Description

Currently there is link http://localhost:3000/ProposalPage which opens list of proposals. 

This should be renamed [http://localhost:3000/proposals ](http://localhost:3000/ProposalPage)

This check is done for all pages to make them inline. 

There is also a fix for showing the logo again


jekabs-karklins (approved):
Looks good

martin-trajanovski:
@jekabs-karklins I see that in the ticket and here in the description is mentioned that links should be like: `http://localhost:3000/proposals` (using small letters only) but this PR still uses links like `http://localhost:3000/Proposals`. Not sure if this is something that works fine or the intention was to use small letters only.

jekabs-karklins:
> @jekabs-karklins I see that in the ticket and here in the description is mentioned that links should be like: `http://localhost:3000/proposals` (using small letters only) but this PR still uses links like `http://localhost:3000/Proposals`. Not sure if this is something that works fine or the intention was to use small letters only.

I noticed that too, but I think as long as it is consistent then it is fine

PR #830, opened 2022-03-24
refactor(formik components): Use existing formik-mui components instead of defining custom ones

martin-trajanovski:
## Description

This PR aims to improve and standardize usage of `Formik` components. Some custom `Formik` components are improved and some are completely removed. We remove some components because there are already defined ones in the `formik-mui` package that we use as a dependency.

Removed components: 
 - FormikUICustomCheckbox (we use equivalent one from formik-mui called: `CheckboxWithLabel`)
 - FormikUICustomSelect (we use equivalent one from formik-mui called: `Select`)

Improved components:
- FormikDropdown (now it is called `FormikUIAutocomplete` we use the formik-mui `Autocomplete` component and it is much more simpler because it includes many functionalities by default)

From now on I think it is very important to make a decision what is better to use in the form from Select or Autocomplete. I think as a general suggestion should be for example if the options are limited up to 10 and fixed list then Select is better choice, otherwise autocomplete.

## Motivation and Context

After upgrading MUI to v5 we had many custom components that are duplicated or not needed anymore.

## How Has This Been Tested

- e2e tests
- lot of manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-2405


src/components/common/FormikUIAutocomplete.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <Field
+      id={name + '-input'}
+      name={name}
+      component={Autocomplete}
+      loading={loading}
+      options={options}
+      noOptionsText={noOptionsText}
+      getOptionLabel={(option: number | string) => {
```
jekabs-karklins, line 42: I think it would make things little cleaner if the method `getOptionLabel` is injected as a prop from the place the component `FormikUIAutocomplete` is used.
This would help f.x. in CallGeneralInfo, where instead of creating `proposalWorkflowOptions` array to be used as items, the items would be simply the property `proposalWorkflows` itself, and there would be just a function that get's the label from that array.

This however could be done in a separate PR, but just to keep in mind

src/components/user/UpdateUserInformation.tsx
```diff
@@ -222,7 +223,7 @@ export default function UpdateUserInformation(props: { id: number }) {
             <Grid container spacing={3}>
               <Grid item xs={6}>
                 <LocalizationProvider dateAdapter={DateAdapter}>
-                  <FormikDropdown
+                  <FormikUIAutocomplete
```
jekabs-karklins, line 226: You should probably use simple select here the same as in SignUps.

If you want to use Autocomplete, there is a parameter for the AutoComplete component `autoComplete` which can be set to false

jekabs-karklins (approved):
Small changes but otherwise looks good

src/components/user/UpdateUserInformation.tsx
martin-trajanovski: Good point!

keiranjprice101 (approved):

src/components/common/FormikUIAutocomplete.tsx
martin-trajanovski: I will create separate task for this and it could be revisited a bit better. Maybe we find some other things like this to expose for improved reusability.

PR #834, opened 2022-03-25
Feat: Add PDF upload to technical review

bolmsten:
## Description

It was requested that an instrument scientist doing a technical review should be able to upload PDF's as part of the technical review. This PR uses the file upload component used in the questionary and adds a file field to the database for technical review.
Screenshot 2022-03-25 at 10 52 38

## Motivation and Context

Requested by Zoe Fisher and approved by User Office

## Fixes

#834 

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/612


jekabs-karklins (approved):
Small comments but looks good

src/components/common/FileUploadComponent.tsx
```diff
@@ -429,7 +432,9 @@ export function FileUploadComponent(props: {
   }
 
   const amountFilesInfo =
-    maxFiles > 1 ? <span>Max: {maxFiles} file(s)</span> : null;
+    maxFiles > 1 ? (
+      <span className={classes.nbrFilesText}>Max: {maxFiles} file(s)</span>
```
martin-trajanovski, line 436: As I reviewed recently some code from @jekabs-karklins and we decided that probably it is a bit better and cleaner to use more the `Box` component that comes from MUI. So this will be much more readable if it is written like: `<Box component="span" display="block">Max: {maxFiles} file(s)</Box>` and remove the custom class that we add in the makeStyles block. This way I think we keep makeStyles smaller and the code is more readable when you see it you don't have to scroll up to see the style definitions.

PR #851, opened 2022-04-07
feat: adding info that there is no guide

jekabs-karklins:
## Description

Adding info that there is no guide if no guide has been set up
image

## Motivation and Context

Currently we show empty page which looks broken

## Changes
GradeGuidePage


src/components/pages/GradeGuidePage.tsx
```diff
@@ -1,16 +1,36 @@
...
 import { useGetPageContent } from 'hooks/admin/useGetPageContent';
 import { StyledContainer } from 'styles/StyledComponents';
 
 const GradeGuidePage: React.FC = () => {
   const [loadingPage, pageContent] = useGetPageContent(PageName.GRADEGUIDEPAGE);
 
-  return (
-    <StyledContainer>{loadingPage ? null : parse(pageContent)}</StyledContainer>
+  const noContents = (
+    <Stack alignItems={'center'}>
```
martin-trajanovski, line 15: This could probably be just `alignItems="center"`.

martin-trajanovski (approved):
Just a small comment otherwise it looks great! I like the usage of `Stack` MUI component!

PR #856, opened 2022-04-11
feat: use the call allocation time unit everywhere where time is used and fix some logic to improve functionality

martin-trajanovski:
## Description

Using the call allocation unit everywhere where it was missing.

## Motivation and Context

Time allocation unit was missing on some places in the SEPs and when you define instrument availability time per call.

## How Has This Been Tested

- e2e tests
- manual testing

## Fixes

https://jira.esss.lu.se/browse/SWAP-2467

## Changes

https://user-images.githubusercontent.com/6333063/163117850-a61de3d8-9acd-40bd-ad2c-b9ea219152bf.mp4

## Depends on

Backend PR: https://github.com/UserOfficeProject/user-office-backend/pull/627


src/components/SEP/MeetingComponents/SEPMeetingComponentsView.tsx
```diff
@@ -28,43 +30,57 @@ const SEPMeetingComponentsView: React.FC<SEPMeetingComponentsViewProps> = ({
...
+          isLoading={loadingCalls}
+          callId={query.call}
+        />
+        <Tooltip title="Export in Excel">
+          <IconButton
+            aria-label="export in excel"
+            className={classes.spacing}
+            data-cy="download-sep-xlsx"
+            onClick={() => {
+              if (query.call) {
```
jekabs-karklins, line 62: Maybe you should disable button instead. If it is clickable and nothing happens that could be confusing

jekabs-karklins (approved):

src/components/SEP/MeetingComponents/SEPMeetingComponentsView.tsx
martin-trajanovski: I see now that the table is disabled anyway when there is no call. So you can't click the button while the table is loading. And also there is no way to remove selection so you must have some call selected.

PR #857, opened 2022-04-12
feat: Adding recommendation column and expanding view

jekabs-karklins:
## Description

This PR adds column to the SEP Meeting Components called Recommendation which shows the recommendation field. This was an idea proposed during walkthrough.

Because extra column was added., the StyledContainer with should be larger. This PR also retrospectively fixes the size of template editor and Proposal view, to make these views larger as well to utilize available screen space
image

## Motivation and Context

As a sep meeting organizer I want to easily see the information in the table about the recommendation of all experiments

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2500

## Changes

Sep meeting components.
StyledContainers for:
- SEP components
- Proposals table
- Template Editor

## Depends on

N/A


src/components/SEP/SEPPage.tsx
```diff
@@ -49,7 +49,7 @@ const SEPPage: React.FC = () => {
   }
 
   return (
-    <StyledContainer>
+    <StyledContainer maxWidth={'xl'}>
```
martin-trajanovski, line 52: This could probably be just maxWidth="xl" without curly brackets.

martin-trajanovski (approved):
Looks good!

src/components/SEP/SEPPage.tsx
```diff
@@ -49,7 +49,7 @@ const SEPPage: React.FC = () => {
   }
 
   return (
-    <StyledContainer maxWidth={'xl'}>
+    <StyledContainer maxWidth='xl'>
```
martin-trajanovski: Isn't this creating an eslint error because usually we use doublequotes for attribute values.

src/components/SEP/SEPPage.tsx
```diff
@@ -49,7 +49,7 @@ const SEPPage: React.FC = () => {
   }
 
   return (
-    <StyledContainer maxWidth='xl'>
+    <StyledContainer maxWidth="xl">
```
martin-trajanovski, line 52: Great I see you fixed it already!

PR #860, opened 2022-04-13
fix: Fix ranking in SEP meeting

jekabs-karklins:
## Description

This PR aims to fix issue where ranking by drag and drop ranked it one position offset. The cause was the dynamic drop area that was placed in the list to indicate the place user can drop the item, but logic was accounting it for an actual item. This caused indexes to be offset.
Peek 2022-04-13 14-15

## Motivation and Context

Because the SEP meeting is in 27 of April we want this feature to be bug free.

## How Has This Been Tested

- [x] E2E
- [x] Manual tests
## Fixes

https://jira.esss.lu.se/browse/SWAP-2588
## Changes

SEPInstrumentProposalsTable.tsx
## Depends on

N/A


martin-trajanovski (approved):
Really nice refactor and improvements.

PR #875, opened 2022-04-26
feat: Use settings formats instead of hardcoding into the initialDBData

martin-trajanovski:
## Description

From now on there is possibility to use database date formats in the e2e tests instead of hardcoding them. The proper use includes calling a function that is defined into the `initialDBData` file `initialDBData.formats()` which returns both `dateTimeFormat` and `dateFormat`. Also before starting a test suite where we want to use this database formats we need to use the function that initializes this database settings: `cy.getAndStoreAppSettings();`

## Motivation and Context

The date formats were hardcoded into `intialDBData` file.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2444

### DIFF
--- a/cypress/support/admin.ts
+++ b/cypress/support/admin.ts
@@ -1,10 +1,25 @@
+import { GetSettingsQuery, PrepareDbMutation } from '../../src/generated/sdk';
 import { getE2EApi } from './utils';
 
-const resetDB = (includeSeeds = false) => {
+const resetDB = (
+  includeSeeds = false
+): Cypress.Chainable<PrepareDbMutation> => {
   const api = getE2EApi();
   const request = api.prepareDB({ includeSeeds });
 
-  cy.wrap(request);
+  return cy.wrap(request);
+};
+
+const getAndStoreAppSettings = (): Cypress.Chainable<GetSettingsQuery> => {
+  const api = getE2EApi();
+  const request = api.getSettings().then((resp) => {
+    window.localStorage.setItem('settings', JSON.stringify(resp.settings));
+
+    return resp;
+  });
+
+  return cy.wrap(request);
 };
 
 Cypress.Commands.add('resetDB', resetDB);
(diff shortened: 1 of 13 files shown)


vyshnavi1605 (approved):

cypress/support/initialDBData.ts
```diff
@@ -1,11 +1,32 @@
...
+} from '../../src/generated/sdk';
 
 // NOTE: Instruments, proposal and scheduled events are seeded only if resetDB(true).
 export default {
-  // TODO: Try to fetch the formats from the DB settings
-  formats: {
-    dateFormat: 'dd-MM-yyyy',
-    dateTimeFormat: 'dd-MM-yyyy HH:mm',
+  // NOTE: To be able to use this cy.getAndStoreAppSettings() should be called in the beforeEach section.
+  formats: () => {
```
jekabs-karklins, line 11: since it is a function I think the better name is getFormats()

cypress/support/admin.ts
```diff
@@ -1,10 +1,25 @@
...
+  includeSeeds = false
+): Cypress.Chainable<PrepareDbMutation> => {
   const api = getE2EApi();
   const request = api.prepareDB({ includeSeeds });
 
-  cy.wrap(request);
+  return cy.wrap(request);
+};
+
+const getAndStoreAppSettings = (): Cypress.Chainable<GetSettingsQuery> => {
```
jekabs-karklins, line 13: If this is something that needs to be run always maybe it is worth looking into support file before hook
https://docs.cypress.io/guides/references/configuration#Folders-Files
https://stackoverflow.com/a/48153074/813773

martin-trajanovski: Will definitely look into that option.

jekabs-karklins (approved):

PR #880, opened 2022-05-03
fix: add copy to clipboar icon

jekabs-karklins:
## Description
Adding icon so it is easier to copy proposal id
image

## Motivation and Context

As a user officer, I want to copy the proposal ID easier

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

N/A

## Depends on

N/A


martin-trajanovski (approved):
Looks nice and reusable.

PR #881, opened 2022-05-03
fix: questions page UX improvements

jekabs-karklins:
## Description

Adding links to relevant places in the site
image
image

## Motivation and Context

As a user officer, I want to access the information quicker

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

N/A

## Depends on

N/A


src/components/questionary/questionaryComponents/QuestionFormShell.tsx
```diff
@@ -28,82 +29,95 @@ const useStyles = makeStyles((theme) => ({
...
-    api()
-      .deleteQuestion({
-        questionId: props.question.id,
-      })
-      .then((data) => {
-        if (data.deleteQuestion.question) {
-          props.onDeleted?.(data.deleteQuestion.question);
-          props.closeMe?.();
+    const submitHandler = async (values: Question): Promise<void> => {
+      api()
```
martin-trajanovski, line 47: I personally think that if you use `async/await` here will be a bit cleaner because we reduce the nesting a bit instead of using `.then` with some additional nesting.

src/components/questionary/questionaryComponents/QuestionFormShell.tsx
```diff
@@ -28,82 +29,95 @@ const useStyles = makeStyles((theme) => ({
...
+            });
+            closeMe?.();
+          }
+        });
+    };
+
+    const deleteHandler = () =>
+      confirm(
+        () =>
+          api()
```
martin-trajanovski, line 68: The same goes here with `async/await` would clean up a bit.

martin-trajanovski (changes requested):

martin-trajanovski (approved):
💯

PR #889, opened 2022-05-05
Create stfc specific cypress config

keiranjprice101:
## Description
Creates STFC specific cypress config to allow running the cypress gui.

I have written these instructions which I plan to add to confluence in the STFC developer workflow section. (formatting not final)

-----------------------------

### Running the tests with the cypress GUI

- 1. Run the frontend with e2e depedency config
- 2. Make sure the frontend is also cloned into windows
- 3. Make sure cypress is installed in windows `npm install cypress --save-dev`
- 4. Run `cypress open -C .\stfc.cypress.json`

From the GUI you can select which specs to run and in which browser. Attempting to run all specs at the same time is not recommended due to the slow speed they are loaded into the cypress GUI.

----------------------------

I would appreciate a review on the documentation as well, before I add it.

## Motivation and Context
Intermission work


martin-trajanovski (approved):
Looks good to me.

sc19ew (approved):
Documentation looks good to me

PR #914, opened 2022-05-12
feat(auth): As a user I want single sign on with PING

jekabs-karklins:
## Description

This PR aims to integrate ESS UserOffice with single-sign-on using the identity access management system PING, developed by our external partners.

## Motivation and Context
The approach for integrating with the external authorization system piggybacks on the external authorization developed by STFC and uses the same approach.

The difference is in UserAuthorization class, which both facilities implement differently.

## How Has This Been Tested

- [x] E2E
- [x] Manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2564

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/657/checks


src/components/user/ExternalAuth.tsx
```diff
@@ -1,31 +1,22 @@
...
+  const externalToken = urlQueryParams.sessionid ?? urlQueryParams.token;
 
-const ExternalAuth: React.FC<ExternalAuthProps> = ({ match }) => {
   const { token, handleLogin } = useContext(UserContext);
   const unauthorizedApi = useUnauthorizedApi();
-  const { search } = useLocation();
-  const values = queryString.parse(search);
-  const sessionId = !!values.sessionid
-    ? values.sessionid.toString()
-    : match.params.sessionId;
```
jekabs-karklins, line 28: @mr-field Could you please check if this change is ok with STFC? It seemed that the {match} parameter wasn't used, so I only left the reading the parameter `sessionId` from the URL

martin-trajanovski (approved):

PR #921, opened 2022-05-19
fix(questionary): fixing missing validation for question template relation

jekabs-karklins:
## Description

This PR addresses the bug where inputs for editing QuestionTemplateRelation were not validated. Instead of validating the values user enters for the `config`, the `default config` object was validated.

## Motivation and Context

The bug prevented the user officer to enter valid values at all, or allowed them to enter wrong values without properly validating

## How Has This Been Tested

- [x] Manual tests
- [x] E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2602

## Changes

QuestionTemplareRelationForms

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/659


cypress/integration/templates.ts
```diff
@@ -1238,6 +1238,29 @@ context('Template tests', () => {
         });
       });
     });
+
+    it('should validate question template relation input', () => {
+      createTopicWithQuestionsAndRelations();
+
+      cy.login('officer');
+      cy.visit('/');
```
martin-trajanovski, line 1246: In scenarios like this I started using this pattern more: `cy.visit('/ProposalTemplates')`. This is one liner that also saves some time in the test.

martin-trajanovski (approved):
In general I think it is quite nice fix and improvement. Just a small comment for better practice maybe.

PR #927, opened 2022-05-20
fix(Technical): The assigned reviewer should be shown as owner of the review

bolmsten:
## Description
At the moment the r"reviewed by" changes based on who last touches the technical review, this behaviour becomes misleading as the user officer cleans up technical reviews and is therefore shown as the person who has done the review. This PR changes this by showing  the assigned reviewer in the "reviewed by" field. 

Screenshot 2022-05-20 at 11 40 08


martin-trajanovski:
@bolmsten The question here probably is what happens if user officer assigns a technical reviewer but officer is the one who does the review instead because he/she has the right to do it. It will still say that technical review is done by the assignee not the officer. Or maybe this is very rare scenario.

bolmsten:
> @bolmsten The question here probably is what happens if user officer assigns a technical reviewer but officer is the one who does the review instead because he/she has the right to do it. It will still say that technical review is done by the assignee not the officer. Or maybe this is very rare scenario.

Yes, I think this is the correct behaviour but we save the last editor still

PR #932, opened 2022-05-23
510 setup mockserver

vyshnavi1605:
## Description

closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/510
## Motivation and Context

Test integration of UOWS
## How Has This Been Tested
Tested by running the relevant E2E tests in STFC mode, and returning mock responses.

## Fixes

fixes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/510
## Changes

- cypress.json 
  - allows the test to navigate to pages outside of the application, this allows stfc logout test to pass
- filterTests.ts 
  - test filter, needed to allow test filtering 
- user.ts and user.d.ts
  - moved changeActiveRole method above externalTokenLogin
  - added externalTokenLogin, if dependency config is stfc the e2e tests will use a this method to login
## Depends on

https://github.com/isisbusapps/docker-orchestration/pull/252
https://github.com/UserOfficeProject/mock-uows/pull/1


sc19ew (approved):

cypress/support/user.ts
```diff
@@ -39,19 +49,86 @@ const testCredentialStore = {
...
+): Cypress.Chainable<LoginMutation | ExternalTokenLoginMutation> => {
   const credentials =
     typeof roleOrCredentials === 'string'
       ? testCredentialStore[roleOrCredentials]
       : roleOrCredentials;
 
+  if (Cypress.env('STFC') === true) {
+    return externalTokenLogin(roleOrCredentials as 'user' | 'officer')
+      .then(() => changeActiveRole(roleOrCredentials === 'user' ? 1 : 2))
+      .then(() => {});
```
mr-field, line 129: Can we remove this `then()`?

cypress/support/user.ts
```diff
@@ -39,19 +49,86 @@ const testCredentialStore = {
...
     | { email: string; password: string }
-): Cypress.Chainable<LoginMutation> => {
+): Cypress.Chainable<LoginMutation | ExternalTokenLoginMutation> => {
   const credentials =
     typeof roleOrCredentials === 'string'
       ? testCredentialStore[roleOrCredentials]
       : roleOrCredentials;
 
+  if (Cypress.env('STFC') === true) {
+    return externalTokenLogin(roleOrCredentials as 'user' | 'officer')
```
mr-field, line 127: I think this would be better handled in one of these 2 ways:
- We can add entries to `testCredentialStoreStfc` to map `user2` and `placeholderUser` to auth tokens, potentially reusing the same ones we already have. We can also check if `roleOrCredentials` is an object and pass `user` to `externalTokenLogin` if so. This would let us support all calls to the `login()` function in STFC mode, as long as the permissions match up
- If we don't want to support tests where `login()` is called with any values other than "user" or "officer", we should check `roleOrCredentials` before calling `externalTokenLogin()` and throw an exception if `roleOrCredentials` isn't "user" or "officer". This will warn us that we tried to run an unsupported test, rather than fail with a less clear reason later on

The main reason I'm asking for this is that the `roleOrCredentials as 'user' | 'officer'` conversion isn't necessarily safe, and it can lead to an error when looking up the token from `testCredentialStoreStfc` if `roleOrCredentials` wasn't one of those values .

cypress/support/user.ts
```diff
@@ -39,19 +55,87 @@ const testCredentialStore = {
...
+    }
+
+    const { user, exp } = jwtDecode(
+      resp.externalTokenLogin.token
+    ) as DecodedTokenData;
+
+    window.localStorage.setItem('token', resp.externalTokenLogin.token);
+    window.localStorage.setItem(
+      'currentRole',
+      roleOrCredentials === 'user' ? 'USER' : 'USER_OFFICER'
```
mr-field, line 105: It would be better to default to the `USER` role if the credentials weren't specifically for a user officer, otherwise all credentials aside from `user` will get user officer permissions:

```suggestion
      roleOrCredentials === 'officer' ? 'USER_OFFICER' : 'USER'
```

mr-field (approved):

martin-trajanovski (approved):
It seems quite good to me.

PR #942, opened 2022-06-01
Stfc fix template import script

keiranjprice101:
## Description

Fixes the issues with the template import script and makes logging clearer.

Requires https://github.com/UserOfficeProject/user-office-backend/pull/666

## Motivation and Context
I previously had only properly checked the delete script and failed to properly check the importScript this now works as expected.

## How Has This Been Tested

Manual Testing

### DIFF
--- a/src/components/experiment/DateFilter.tsx
+++ b/src/components/experiment/DateFilter.tsx
@@ -1,6 +1,6 @@
 import { DatePicker, LocalizationProvider } from '@mui/lab';
 import DateAdapter from '@mui/lab/AdapterLuxon';
-import { Stack, TextField } from '@mui/material';
+import { Stack, TextField, TextFieldProps } from '@mui/material';
 import { makeStyles } from '@mui/styles';
 import { DateTime } from 'luxon';
 import React from 'react';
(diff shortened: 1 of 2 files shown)


scripts/importTemplates.js
```diff
@@ -117,55 +105,91 @@ async function main() {
...
 
     console.log('Template import result:');
     console.log(importTemplateResult.data.data);
-  });
+    console.log('------------------------------------------------------------');
+  }
 }
 
-main();
+main().catch((e) => console.log(e));
```
jekabs-karklins, line 195: This could call console.error()

jekabs-karklins (approved):

scripts/importTemplates.js
```diff
@@ -117,55 +105,91 @@ async function main() {
...
 
     console.log('Template import result:');
     console.log(importTemplateResult.data.data);
-  });
+    console.log('------------------------------------------------------------');
+  }
 }
 
-main();
+main().catch((e) => console.error(e));
```
ACLay, line 195: There's a couple of other promise `.catch` methods that call `console.log` Should they be changed to `error` too?

ACLay (commented):
Maybe it's just a lack of familiarity with the project/structure, but I can't figure out where or how this script is called or used. Could you give me a quick run through? Would that be worth adding as comments at the start of the script?

keiranjprice101:
@ACLay The intention is for this script and the delete template script to be run periodically by jenkins to keep dev and prod templates in sync. That is why its not called anywhere else. I will add a comment to make this more clear 👍

vyshnavi1605 (approved):

PR #947, opened 2022-06-08
feat: Add page limit for PDF file uploads

simonfernandes:
## Description

This allows you to specify a PDF page limit for file upload questions. It has a default value of 0 (unlimited pages) and the option to modify this only shows when PDF is one of the accepted file types for the question. 

On the backend there is some validation using the muhammara library to parse the PDF and check the number of PDF pages detected against the specified limit.

image

image

## Motivation and Context

Facilities at STFC have different PDF page limitations for certain files.

## How Has This Been Tested

- Tested file submission via questionnaire locally
- Tested file submission via technical review locally
- Added e2e tests

## Fixes

Fixes UserOfficeProject/user-office-project-issue-tracker#19

## Changes

- Show new PDF page limit option when creating PDF file upload questions
- Add backend validation to check number of PDF pages
- Do the different backend file validations sequentially and fail early (should be a bit quicker)

## Depends on

Backend PR: https://github.com/UserOfficeProject/user-office-backend/pull/675


simonfernandes:
I'm struggling to get the e2e tests passing on this. Most of the failures seem to be unrelated. Going to put this up for review for now and see if they resolve themselves or fix them up later

mr-field:
@simonfernandes The E2E tests and frontend build should be fixed now, merging in `develop` should take care of the problems.

src/components/common/FileUploadComponent.tsx
```diff
@@ -366,6 +366,7 @@ export function FileUploadComponent(props: {
   maxFiles?: number;
   id?: string;
   fileType: string;
+  pdfPageLimit: number;
```
martin-trajanovski, line 369: Maybe it is a good idea to have a comment here that `0` means unlimited pages. Because I see we use the component in couple of places and would be nice in the future if we want to add to another place to know that we should set this property to 0 for unlimited.

martin-trajanovski (approved):
Just a small comment otherwise it looks quite nice!

src/components/common/FileUploadComponent.tsx
jekabs-karklins: Would also be nice to let the user know it somehow

jekabs-karklins (approved):

src/components/common/FileUploadComponent.tsx
simonfernandes: Thanks both 🙂

@jekabs-karklins I think this exists at the moment, unless you meant something more?

image

image

mr-field (approved):

PR #953, opened 2022-06-13
feat: add cypress test filter for ess and stfc tests

vyshnavi1605:
## Description

Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/511

Best reviewed with whitespace off: https://github.com/UserOfficeProject/user-office-frontend/pull/953/files?diff=split&w=1
## Motivation and Context

Some tests will not pass in the stfc environment due to lack of features, so adding a filter allows us to choose what tests we want to run in each environment 
## How Has This Been Tested

Tested by running the tests with specific features enabled and disabled
## Fixes

https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/511
## Changes

introduced testing based on featured enabled 
Called `cy.getAndStoreFeaturesEnabled(); ` in `beforeEach` in every spec file to get the enabled features. The method for this is in `user.ts`  mapping the enabled features logic is in `featureFlags.ts`. The method had to be in `user.ts`, was initially in `admin.ts` but localstorage was not storing the contents in that file. 

Another beforeEach was required in addition to the existing one, to get the correct features enabled. This was needed to allow the right tests to get skipped. 

- calls - **(2 tests have an if statement within the test so that part does not execute yet testing the over test)**
- eventLog - **(1 test uses USER_MANAGEMENT flag)** 
- instruments - **(last describe block involve logging in a scientist, logging in as scientist and logging in with email and password doesn't work for stfc login. This can be changed with an if statement directing to the right login. But mock responses don't seem to be working right now, so until they work this describe block will not execute if EXTERNAL_AUTH is enabled.)**
- peopleTable - **(mock responses don't seem to be working right now, so until they work these tests will not execute if EXTERNAL_AUTH is enabled. This will be removed after mock responses are fixed)**
- samples -  **(tests use SAMPLE_SAFETY flag will only run if that flag is enabled)** 
- settings - **(some tests need feature flags, so those tests only ruin if the relevant flag is enabled)** 
- templates - **(some tests need feature flags, so those tests only ruin if the relevant flag is enabled)** 
   - User officer can delete active template seems to be getting skipped on github if any other feature flag other than external auth. We don't run that test is external auth is enabled anyway. But the actual feature flag for that is shipment 
- some spec files test certain features which are behind feature flags and they will only execute if the relevant flag is enabled. 
allow tests to be carried out the ess environment.  
## Depends on

https://github.com/isisbusapps/docker-orchestration/pull/291
https://github.com/UserOfficeProject/mock-uows/pull/4

- [ ] All relevant doc has been updated - once these changes are merged in I will update the confluence page


vyshnavi1605:
> What does the `CYPRESS_STFC` variable do?

It is for checking if the tests should run in the stfc environment or the ess environment

ACLay (approved):

jekabs-karklins:
@vyshnavi1605 could you please elaborate a little on why are tests failing in the STFC environment?

vyshnavi1605:
> @vyshnavi1605 could you please elaborate a little on why are tests failing in the STFC environment?

Sure, 
- In the calls and proposals tests a part of the test involves an esi template which is behind the feature flag therefore not accessible in stfc environment. I briefly remember hearing that we can enable certain features in tests so I will need to look into that. 

- The tests in the peopleTable are failing as the creation of users in the before each is failing, I think its failing at this point: `//Check if email exist in database and if user has been invited` but I need to look into it further

- For settings and templates, I need to write the accurate mock responses based on tests and debug why they are failing currently in the stfc environment

mr-field (commented):
I think an important change from this would be that we have to add `TestFilter` to all Cypress tests, bit now and in the future.

Is there any way of deciding which tests we run somewhere else? For example, I think you can define which tests you want to run in the `cypress.json` file, or as arguments to the `cypress` command. Could we have an STFC-specific and an ESS-specific JSON file which sets which tests run in each environment? I think it would make it easier to see what's running where and it would remove the need to change the code at all.

jekabs-karklins:
Would it be possible instead of disabling tests, to write another one to test the feature with different feature flags enabled? 
So there would be a test called "Create call creation with RISK_ASSESSMENT disabled", which would disable the feature and test if the flow. 
I think in that way we would also always test the flow when certain features are disabled and make sure the feature works in different configurations. Let's talk about it during the next meeting.

vyshnavi1605:
> I think an important change from this would be that we have to add `TestFilter` to all Cypress tests, bit now and in the future.
> 
> Is there any way of deciding which tests we run somewhere else? For example, I think you can define which tests you want to run in the `cypress.json` file, or as arguments to the `cypress` command. Could we have an STFC-specific and an ESS-specific JSON file which sets which tests run in each environment? I think it would make it easier to see what's running where and it would remove the need to change the code at all.

I tried adding the spec files I want to run the cypress command, and it does work. However, in the spec files we run run selective tests based on environment. E.g in `eventLogs.ts` there are two tests and only one of them is applicable in the stfc environment. I can't seem to find a way of setting only specific `it/describe` blocks through the command. 

Another option I came across is to folderise the tests, we can point multiple folders to run at once. Maybe something like:
├── cypress
│    ├── integration
│    │    ├── stfc
│    │    ├── ess
│    │    ├── stfcEss

stfc will have only stfc specific tests and will the ess folder. And in the stfcEss folder we will add the tests that will run in both environments. The test files can be named the same if we wanted to, for e.g `eventlogs.ts` the test that will run in both stfc and ess environment will be in the `eventLogs.ts` file in the `stfcEss` folder and the test that is only specific to the ess will be in a file called `eventLogs.ts` in the ess folder. 


Tests in the `calls.ts` file testing call creation, will need to also be ran in the stfc environment by disabling the RISK_ASSESMENT as Jekabs said. Only tests that do not need to be ran in the stfc environment will go in the ess folder. 

The commands will be: `--spec cypress/integration/stfcEss/*.ts cypress/integration/ess/*.ts` and `--spec cypress/integration/stfcEss/*.ts cypress/integration/stfc/*.ts`

This change will mean that when new tests are added it has to be added and separated out into the correct folders to be executed.

vyshnavi1605:
After the sprint planning meeting on Wednesday, we agreed to firstly test out and check why the tests are failing in the stfc environment. 

Falling tests: 
- eventLogs.ts 
    - **If user uptates his info, officer should be able to see the event logs for that update** failing point: ` cy.contains('People').click();` after enabling the feature flag, appropriate mock responses need to be added by checking the parameters. 

- experiments.ts 
    - need to enable feature flag
    - **Can view visits** test fails, failing point: ```
cy.get('[data-cy=officer-scheduled-events-table]').contains(
        initialDBData.users.user1.lastName)```. Seems to be getting the wrong user, need to check why its getting the wrong response get make sure it gets the right mock response 

- generalSeps.ts 
       - need enable feature flag
       - majority are still failing as incorrect responses are returned SEP responses also don't exist

- institutions.ts 
       - need to enable feature flag
       - **User Officer should be able to merge Institution** failing as it calls this method in the web service `getBasicPeopleDetailsFromUserNumbers` and this mock response needs altering and fixing. 

- instruments.ts 
      - enable technical review feature flag for some tests, 
      - failing at login, logging in as `scientist` and logging in with email and password doesn't work for stfc login. 

- peopleTable.ts 
      - Trouble creating users, creating the users with user id from mock response. Potentially something wrong with the mock response. 

- personal_information.ts 
       - need to enable feature
       - problematic mock response, SEP Chair role doesn't exist in stfc user

- proposal_adminstration 
      - **Should be able to re-open proposal for submission** failing at logout
      - feature flag needs to be enabled for 3 tests and they pass - 
	- **User officer should see Reviews tab before doing the Admin(management decision)**
	- **If you select a tab in tabular view and reload the page it should stay on specific selected tab**
	- **Should be able to set comment for user/manager and final status**

- proposal_esi 
     - failing at login, logging in as PI and logging in with email and password doesn't work for stfc login

- scheduler.ts
      - enable feature flag
      - tests that include logging out are failing

- SEP.ts 
      - User officer role, some mock responses are incorrect. 
      - SEP chair role, doesn't exist 

- shipment.ts 
      - enable feature flag
      - login failed with email and password
      - tests that include logging out are failing

- user_adminstration.ts 
          - enable feature flag 
          - tests that include logging out are failing
          - /signIn takes it to dashboard
          - email to reset password test failing 

- user_spec.ts 
      - logout and login test fails

- visits.ts
     -  enable feature flag
     -  login with email + password test failing

- settings.ts 
   -  enable feature flags 
    -  **Proposal should follow multi-column workflow** - failed at logout
    -  **Proposal status should update immediately after all SEP reviews submitted** - mock responses incorrect
    -  **Proposal should follow the selected workflow** - failed at logout

- templates.ts
    -  logout causing test failure. 


Questions:
- Most of these test files apart from, `settings.ts`, `templates.ts`, `instruments.ts`, `proposal_administration.ts` and `peopleTable.ts` seem to be testing features that are not used and are hidden in the stfc side. Would it still be needed to run these tests in the stfc side and debug, figure out and write appropriate mock responses? Or would adding a filter to filter those tests to be ran in the stfc side be an appropriate option?

mr-field:
>  Would it still be needed to run these tests in the stfc side and debug, figure out and write appropriate mock responses (I think this might involve good amount of time) or would adding a filter to filter those tests that test hidden features in the stfc side be an appropriate option as we do want this before direct?

I think we want to avoid assigning tests to a specific facilities, because many are shared and we want to run them everywhere. It's also a problem if more facilities join the collaboration, as they will need their own testing config too. It's also good to avoid breaking up the app into "this bit is only for this facility", as in theory everything can be used by everyone.

Instead of filtering tests by facility, maybe we could change the filter to look at the currently enabled feature flags, and only runs a test suite if the relevant flag is enabled? That way we would only test the features which the instance we're currently testing claims to support. We could then run the tests twice, once with the ESS config and once with the STFC config, and in each case only the set of features used by a certain facility would be tested.

For example, we could rename the test filter to `IfFeaturesEnabled`, making it take a list of feature flags rather than facility names. The filter would then get the currently enabled features, which it should be able to do by querying the backend. If all the specified features are enabled, it runs the test; otherwise, it skips it. This would allow STFC to avoid testing features which aren't (yet) integrated with our other tools and are currently disabled, while still testing all the ones we're actually using.

We could also add an `IfFeaturesDisabled` filer, which does the reverse (only run a test if a feature is diabled). This would let us skip tests which test functionality that doesn't make sense with external identity management- for example, we can't test if users details were correctly updated in the STFC setup because user details aren't managed though the app.

vyshnavi1605:
Updated initial PR comment to explain my changes

cypress/integration/SEP.ts
```diff
@@ -137,6 +139,7 @@ function createWorkflowAndEsiTemplate() {
 }
 
 function initializationBeforeTests() {
+  cy.getAndStoreFeaturesEnabled();
```
jekabs-karklins, line 142: Do you think we could move `cy.getAndStoreFeaturesEnabled();` into cypress/support/beforeRunHook.ts so that it is run always before every test, and we can clean up these before hooks in each test?

It does not hurt if it is not used, right?

jekabs-karklins (approved):

cypress/integration/SEP.ts
vyshnavi1605: Just putting into the cypress/support/beforeRunHook.ts doesn't seem to work as it doesn't get the features enabled and just returns an empty object. I think `cy.getAndStoreFeaturesEnabled()`; needs to be in each of the beforeEach.

jekabs-karklins: Yes, I see that the hook is run before the setup of the DB.

jekabs-karklins:
I think we can merge this now

ZenixTheHusky (approved):
Looks good to me :)

cypress/integration/proposal_administration.ts
```diff
@@ -162,7 +168,11 @@ context('Proposal administration tests', () => {
         );
     });
 
-    it('Should be able to re-open proposal for submission', () => {
+    it('Should be able to re-open proposal for submission', function () {
+      // if (!featureFlags.getEnabledFeatures().get(FeatureId.EXTERNAL_AUTH)) {
+      //   //temporaily disable test in for stfc until logout function works
+      //   this.skip();
+      // }
```
ZenixTheHusky: Spoken to @vyshnavi1605 about these commented lines, will re-review once these are removed

jekabs-karklins (approved):

ZenixTheHusky (approved):

PR #964, opened 2022-07-06
feat: Add questionary preview for proposal templates

martin-trajanovski:
## Description

This adds the possibility for the user officer to be able to see the questionaries' while building them. For now it works only with Proposal templates but could be extended for other templates as well.

## Motivation and Context

Previously there was no way for a user officer to preview the questionaries' that were created. You need to login as a user and start creating proposal to be able to see it how it really looks like.

## How Has This Been Tested

- e2e tests
- manual tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2425

## Changes

https://user-images.githubusercontent.com/6333063/180430751-688974d9-174e-46aa-b2c4-e03d8f4c728d.mp4

### DIFF
--- a/cypress/integration/templates.ts
+++ b/cypress/integration/templates.ts
@@ -376,7 +376,7 @@ context('Template tests', () => {
       cy.contains(newName).should('not.exist');
     });
 
-    it('User officer can modify proposal template', () => {
+    it('User officer can modify and preview proposal template', () => {
       cy.login('officer');
       cy.visit('/');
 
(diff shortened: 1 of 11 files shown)


cypress/integration/templates.ts
```diff
@@ -461,6 +461,30 @@ context('Template tests', () => {
...
+      /* Check if template preview works */
+      cy.get('[data-cy="preview-questionary-template"]').click();
+      cy.get('[aria-labelledby="preview-questionary-template-modal"]').should(
+        'exist'
+      );
+
+      cy.get('[data-cy="questionary-stepper"] button').last().click();
+
+      cy.get(
+        '[aria-labelledby="preview-questionary-template-modal"] form'
```
jekabs-karklins, line 473: I wonder if we can use data-cy property instead of aria-labelledby or if you think using aria-* labels is fine. Not sure.

jekabs-karklins (approved):

cypress/integration/templates.ts
martin-trajanovski: Using data-cy property on the Dialog component is not working. The only solution to this is to wrap it in a `div` or some other proper element and add data-cy there (which to be honest I don't like much). I would prefer to keep it like this for now and if there is some general solution to all modals to be able to add `data-cy` prop would be nice for later usage.

PR #971, opened 2022-07-26
626 arrows on mcq question duplicate selected option

mutambaraf:
## Description

Closes:https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/626

## Motivation and Context

The arrows in the item tile in the MCQ do not seem to work. The expected functionality is that when you have two or more options, using the arrows you can decide what order you want them to show. Currently when you select the arrow of the option you want to move up or down, the option that was previously there changes its value.

## How Has This Been Tested

Is coved by  integration questions test

## Fixes

Fix : https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/626
## Changes

   Files changed: 
    FormikUICustomTable.tsx
    cypress/integration/templates.ts

### DIFF
--- a/src/utils/helperFunctions.ts
+++ b/src/utils/helperFunctions.ts
@@ -120,3 +120,30 @@ export const removeColumns = <T extends object>(
     }
   });
 };
+
+export const isObject = (object: null) => {
+  return object != null && typeof object === 'object';
+};
+export const isObjectsEqual = (obja: object, objb: object) => {
+  if (!(obja && objb)) {
+    return false;
+  }
+  const propsa = Object.getOwnPropertyNames(obja);
+  const propsb = Object.getOwnPropertyNames(objb);
+  if (propsa.length != propsb.length) {
+    return false;
+  }
+  for (let i = 0; i < propsa.length; i++) {
+    const vala = obja[propsa[i] as keyof typeof obja];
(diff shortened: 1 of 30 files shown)


src/utils/helperFunctions.ts
```diff
@@ -120,3 +120,30 @@ export const removeColumns = <T extends object>(
     }
   });
 };
+
+export const isObject = (object: null) => {
+  return object != null && typeof object === 'object';
+};
+export const isObjectsEqual = (obja: object, objb: object) => {
```
martin-trajanovski, line 127: I see in the code `utils/json.ts` there is one function called `deepEqual`. Can you please check if it does the same comparison for object equality and use that one instead or if you think this one is better then swap that one with yours.

martin-trajanovski (changes requested):
Functionality seems good to me but would be good to have an e2e test if possible to cover this bug.

src/utils/helperFunctions.ts
mutambaraf: They both perform the same task l want  equally but deepEqual covers more object types ,so l be using it instead of isObjectsEqual

martin-trajanovski (approved):

vyshnavi1605 (approved):

PR #977, opened 2022-08-04
feat(auth): Implement openid connect authorizer

jekabs-karklins:
## Description

This PR aims to add support for authentification with external identity providers using OpenId connect.

## Motivation and Context

As we are integrating with the Ping Identity Access Management platform which supports OpenId Connect which is a simple identity layer on top of the OAuth 2.0 protocol, the implementation has been made generic, so that it is possible to use any other Identity Provider platform implementing the OpenId connect.

This PR contains also an example of Google Authorization as can be seen from the example.development.env in the backend PR

IMPORTANT:
In the development, environment authorization is performed with a locally run simple OAUTH server (for more information inspect updated docker.compose file) in the backend
after this PR is merged if the OAuth authorizer is used. then new .env values must be set

## How Has This Been Tested

- [x] Manual tests
- [x]  E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2688

## Changes

Deleted Ping specific code and started using OAuth flow with code response type
Improved error handling and prevented infinite redirect loop if an error occurred.

## Depends on
https://github.com/UserOfficeProject/user-office-backend/pull/707


src/components/common/CenteredAlert.tsx
```diff
@@ -0,0 +1,40 @@
...
+    },
+  },
+}));
+
+/** Message occupying full screen and displaying slightly customized MUI Alert component */
+function CenteredAlert(props: AlertProps) {
+  const classes = useStyles();
+
+  return (
+    <Grid className={classes.grid}>
```
martin-trajanovski, line 34: Maybe it is better to use the `Box` component here with its properties instead of using Grid with custom styles. If you have some time try to do it that way.

src/context/UserContextProvider.tsx
```diff
@@ -141,11 +145,7 @@ const reducer = (
         currentRole: action.payload.toUpperCase(),
       };
     case ActionType.LOGOFFUSER:
-      localStorage.removeItem('token');
-      localStorage.removeItem('currentRole');
-      localStorage.removeItem('user');
-      localStorage.removeItem('expToken');
-      localStorage.removeItem('impersonatingUserId');
+      clearSession();
```
martin-trajanovski, line 148: Nice helper method!

martin-trajanovski (approved):
Looks good to me! Just some comments left to check.

src/components/App.tsx
```diff
@@ -106,8 +126,6 @@ const Routes: React.FC<RouteProps> = () => {
     return (
       <div className="App">
         <Switch>
-          <Route path="/SignUp" component={SignUp} />
-          <Route path="/SignIn" component={SignIn} />
```
ACLay, line 110: It looks to me like this PR is removing the routing and component for `/SignIn`, but line 92 in this file still has a redirect to that path. Should that have been updated?

src/utils/clearSession.ts
```diff
@@ -0,0 +1,9 @@
+const clearSession = () => {
+  localStorage.removeItem('token');
+  localStorage.removeItem('currentRole');
+  localStorage.removeItem('user');
+  localStorage.removeItem('expToken');
+  localStorage.removeItem('impersonatingUserId');
+};
```
ACLay, line 7: Is there anything put into localStorage that we do want to persist between sessions? If there aren't, could this be replaced with `localStorage.clear()`?

src/components/common/CenteredAlert.tsx
jekabs-karklins: I tried that initially but because of more advanced CSS rules, this looked cleaner.

src/components/App.tsx
jekabs-karklins: I have removed the reference now. I also removed EXTERNAL_AUTH feature flag because all auth is now external and not part of UserOffice

src/utils/clearSession.ts
jekabs-karklins: Although clear(), would also work, for now, I think it's safer to remove the specific items.

ACLay: I don't feel that manual removal is safer. If someone in the future copies how we set a session variable without being aware of the need to clear it that could lead to bugs. Using `clear()` eliminates that risk.

jekabs-karklins: Yes, I see your point and had exactly the same considerations. What I also considered is that localStorage could contain other data not related to sessions that some features might want to persist. For example right now if clear() was added it would alter the functionality of the drawer, whose state of open/close is persisted in localstorage even if you log out and back in again. We can have a small discussion in the meeting. Another option would be to implement some `SessionManager` that contains the logic

src/components/user/SharedAuth.tsx
```diff
@@ -38,47 +25,30 @@
...
-      <div className={classes.root}>
-        <UOLoader />
-        <div className={classes.message}>Please wait</div>
-      </div>
-    );
+      return null;
+    }
   }
 
+  window.location.href = authRedirect as string;
```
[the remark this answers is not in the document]
jekabs-karklins: This is a temporary solution because the scheduler should use OAuth too instead of this SharedAuth.

cypress/support/user.ts
```diff
@@ -179,9 +179,13 @@ const login = (
...
 const logout = () => {
-  cy.get('[data-cy=profile-page-btn]').click();
+  localStorage.removeItem('token');
+  localStorage.removeItem('currentRole');
+  localStorage.removeItem('user');
+  localStorage.removeItem('expToken');
+  localStorage.removeItem('impersonatingUserId');
 
-  cy.get('[data-cy=logout]').click();
+  cy.visit('/');
```
ACLay, line 188: Why does this change the tests to manually clear out the session variables? I'd have thought clicking logout would mirror the way a user does it, and mean we don't need to keep the two lists of session items in sync

src/utils/clearSession.ts
ACLay: Can we add a comment to `clearSession` explaining this so future devs can see more easily what should be listed in here and why we don't just use `localStorage.clear()`

ACLay (changes requested):
Is it possible to do a simultaneous merge between this and your scheduler so we don't have to add the workaround that github's code scanner flagged with "Client-side cross-site scripting" and "Client-side URL redirect" warnings?

jekabs-karklins:
> Is it possible to do a simultaneous merge between this and your scheduler so we don't have to add the workaround that github's code scanner flagged with "Client-side cross-site scripting" and "Client-side URL redirect" warnings?

Yes, good idea. It will make PR little bigger, but then it's done in one go

src/utils/clearSession.ts
jekabs-karklins: Done

jekabs-karklins:
> Is it possible to do a simultaneous merge between this and your scheduler so we don't have to add the workaround that github's code scanner flagged with "Client-side cross-site scripting" and "Client-side URL redirect" warnings?

Done

cypress/integration/instruments.ts
```diff
@@ -68,7 +68,7 @@ context('Instrument tests', () => {
     });
 
     it('User officer should be able to create instrument', function () {
-      if (featureFlags.getEnabledFeatures().get(FeatureId.EXTERNAL_AUTH)) {
+      if (!featureFlags.getEnabledFeatures().get(FeatureId.SCHEDULER)) {
```
ACLay, line 71: There's a couple of tests here that are changed to skip if the scheduler isn't used. We don't use the scheduler, but do have instruments. I wonder if these skip's should be removed rather than updated?

src/components/user/ExternalAuth.tsx
```diff
@@ -9,53 +15,156 @@ import { useUnauthorizedApi } from 'hooks/common/useDataApi';
...
+
+    const handleCode = (code: string) => {
+      const { protocol, host, pathname } = window.location;
+      const currentUrlWithoutParams = [protocol, '//', host, pathname].join('');
+
+      setView(<ContactingAuthorizationServerMessage />);
+
+      unauthorizedApi()
+        .externalTokenLogin({
+          externalToken: code,
```
ACLay: Given that it seems `code` is only used once, in `handleCode` to set `externalToken`, could we rename it to `token` through this file? I think that'd be easier to follow, as reading `code` my mind keeps expecting to mean some javascript or an http status code, and there's no comments saying what `code` should be

ACLay (changes requested):

src/components/user/ExternalAuth.tsx
jekabs-karklins: Yes, name `code` is quiet generic. I will change it to `authorizationCode`. The token would be a bad name for OAuth flow, because `token` in OAuth is a different thing. `code` short lived string that is is exchanged for for `token`. check this image https://darutk.medium.com/diagrams-and-movies-of-all-the-oauth-2-0-flows-194f3c3ade85 step 6.

cypress/integration/instruments.ts
jekabs-karklins: I agree, but enabling these tests would be a different PR. I have a suspicion that they have been disabled because they were failing for some other reason

PR #983, opened 2022-08-10
refactor(graphql): Split getFields into countries and nationalities queries

jekabs-karklins:
## Description
This PR splits ambiguous endpoint getFields into nationalities and countries graphQL queries. Not only does it remove unnecessary `get` in from of query which makes it consistent with other queries, but also splits the endpoint into two separate calls because they are unrelated.

## Motivation and Context
This PR aims to make GraphQL endpoints more consistent and more friendly for 3rd party integration

## How Has This Been Tested

- [x] Manual tests
- [x] E2E tests

## Fixes

https://jira.esss.lu.se/browse/SWAP-2729

## Depends on

https://github.com/UserOfficeProject/user-office-backend/pull/708


src/hooks/user/useGetNationalities.ts
```diff
@@ -0,0 +1,33 @@
+import { useEffect, useState } from 'react';
```
martin-trajanovski, line 1: This seems fine but I am concerned about the naming of the hooks and the files in general. For example I see that everywhere else we use this pattern `useUserData` or just `useFeatures` but now you are including this `Get` keyword in the naming of the hooks and files. Not sure if this is necessary because we use the hooks to query something and it is quite understandable that it is for getting some data. You can probably use just `useNationalities` and `useCountries`. Let me know what do you think about this. Just mentioning because I want to keep the consistency.

martin-trajanovski (changes requested):
The refactoring seems nice. There is a small change request from my side in one of the comments. Please take a look and I can review again.

src/hooks/user/useGetNationalities.ts
jekabs-karklins: These of course should be called like that

martin-trajanovski (approved):
This looks great now

TickleThePanda (approved):

PR #988, opened 2022-08-12
feat: Add PDF templates

simonfernandes:
## Description

This is a PR that simply creates a new PDF template type. User officers can go into a new PDF template section and use the same controls that exist for other template types (create, edit, delete, archive). There is a simple page with a single input for the user officer to put in template contents. 

Eventually, the PDF template can be attached to the call (not in this PR), and when sent to the factory will render the PDF using the template defined.

Behind the scenes, the PDF type follows how existing templates have been done, with a few small differences since PDF templates are non-questionary. In order to re-use a lot of the existing template structure in the backend and components in the frontend, there are a few conditionals in the backend where PDF templates are handled slightly differently, e.g. in their creation. In the frontend, I have tried to re-use existing template components where possible, but have had to create a new (and simplistic) PDF Editor for example. To make use of more of the existing template components would require more complex refactoring.

A few things don't yet work: templates can't be deleted, and a call count should display alongside each template, and the title and desription is not editable after creation. If this isn't yet polished enough for the ESS, I could hide this behind a feature flag for now.

## Motivation and Context

We would like to sometimes render a proposal PDF using a custom template.

## How Has This Been Tested

- Unit tests and e2e tests
- Checked template data is created in table(s)

## Fixes

Part of UserOfficeProject/stfc-user-office-project#613

## Changes

Mainly new additions for PDF type. A few tweaks to Template type and components to accommodate.

## Depends on

Backend: https://github.com/UserOfficeProject/user-office-backend/pull/716


src/components/template/PdfTemplatesTable.tsx
```diff
@@ -0,0 +1,35 @@
...
+    >[]
+  >;
+};
+
+const columns: Column<TemplateRowDataType>[] = [
+  { title: 'Name', field: 'name' },
+  { title: 'Description', field: 'description' },
+];
+
+function PdfTemplatesTable(props: PdfTemplatesTableProps) {
```
mr-field, line 22: I'm not that good at React, but does this component ever get used? Seems like the sidebar link goes to `PdfTemplatesPage`, which includes `DefaultTemplatesTable` rather than this table

src/components/template/PdfTemplateEditor.tsx
```diff
@@ -0,0 +1,92 @@
...
+import useDataApiWithFeedback from 'utils/useDataApiWithFeedback';
+
+export default function PdfTemplateEditor() {
+  const { api } = useDataApiWithFeedback();
+  const [loading, setLoading] = useState<boolean>(true);
+  const [template, setTemplate] = useState<Template | null>(null);
+  const [pdfTemplate, setPdfTemplate] = useState<PdfTemplate | null>(null);
+
+  // TODO: get the template ID a better way
+  const templateId = Number(window.location.pathname.split('/').pop());
```
mr-field, line 22: Could this be passed as a React prop somewhere?

jekabs-karklins: To get it from URL I think this would work
```
import { useParams } from 'react-router';
const { pdfTemplateId } = useParams<{ pdfTemplateId: string }>();
```

jekabs-karklins (approved):

src/components/template/PdfTemplatesTable.tsx
simonfernandes: Thanks, it was unused! I made it when I thought I was going to have to create my own and forgot to remove it :)

src/components/template/PdfTemplateEditor.tsx
simonfernandes: Great to know, thanks!

.github/workflows/test-build.yml
```diff
@@ -24,24 +24,34 @@ jobs:
...
-              # to track what we found
-              >&2 echo "$endpoint -> $branch_found"
-              if [[ $branch_found != "" ]]; then
-                found=$((found += 1))
-              fi
-            done
+
+            local dockerhub_endpoint=$1
+            local dockerhub_json=$(curl -s $dockerhub_endpoint)
+            local dockerhub_branch_found=$(echo $dockerhub_json | jq -c ".results[] | select(.name==\"$branch\") | .name")
```
TickleThePanda, line 30: Sorry this change is tacked onto this - wanted to make sure this PR worked while DockerHub was in brownouts..

Dockerhub's v1 API [is deprecated and will be switched off on 5th September](https://www.docker.com/blog/docker-hub-v1-api-deprecation/). They're currently doing [brownouts](https://twitter.com/Docker/status/1561707850543632385).

src/components/template/PdfTemplateEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+                label="Template contents"
+                type="text"
+                component={TextField}
+                data-cy="template-data"
+                fullWidth
+                multiline
+                minRows={10}
+                variant="outlined"
+                onChange={handleChange}
+                value={values.templateData}
```
jekabs-karklins, line 82: If for the component you use  TextField imported from `import { TextField } from 'formik-mui';` then this line is not needed.

jekabs-karklins (approved):

mutambaraf (approved):

src/components/template/PdfTemplateEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+                  );
+                });
+            }
+          }}
+        >
+          {({ handleChange, values }) => (
+            <Form>
+              <Field
+                name="templateData"
+                id="templateData"
```
mutambaraf, line 73: We can also make the field name and id to have different names if possible.

TickleThePanda: I think in many areas of the application we do the same as this here - I'm going to merge for now and we can discuss this further at continuous improvement.

PR #990, opened 2022-08-16
creating github actions for e2e tests to run in stfc environment

vyshnavi1605:
## Description

Closes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/512
## Motivation and Context
Allow e2e tests to run in stfc environment 

## How Has This Been Tested
Tested by running e2e tests on GitHub

## Fixes
Fixes: https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/512

## Changes

- Changes to: `test-build.yml` and added new file: `docker-compose-stfc.e2e.yml`

- `Instrument Scientist filter should differ based on setting value` in appSettings.ts is disabled if extrernalAuth is enabled, this is temporary until instrument scientist role is enabled.  

- `Should be able to download proposal pdf` in proposal_administration is temporarily disabled if extrernalAuth is enabled until test passes in stfc config , this test runs and works fine locally in stfc and e2e config. But does not work in stfc config on GitHub. A 500 Internal Error is returned. 

- Made the `logLevel ERROR` as `INFO` involved a lot of logs and they got truncated, therefore if some tests failed in stfc config we couldn't see which ones. 

**NOTE:** This now means there will be a slight change that we need to add when we add new tests. I will update and document that in the confluence page. 

Will open issues to create and fix currently disabled failing tests in stfc config


.github/workflows/test-build.yml
```diff
@@ -329,6 +329,37 @@ jobs:
...
+          export USER_OFFICE_BACKEND=http://backend:4000/graphql
+          export USER_OFFICE_ENDPOINT=$USER_OFFICE_BACKEND
+
+          export USER_OFFICE_FACTORY_TAG=${{ needs.resolve_dep.outputs.FACTORY_TAG }}
+          export USER_OFFICE_FACTORY_ENDPOINT=http://factory:4500/generate
+
+          export USER_OFFICE_FRONTEND_DIR=$REPO_DIR_NAME
+          export BUILD_TAG=${{ github.sha }}
+
+          export CYPRES_SPEC_PATTERN=${{ matrix.pattern }}
```
keiranjprice101, line 352: Is this meant to be cypres instead of cypress?

vyshnavi1605: We call it in the `docker-compose.yml` file like like: `cypress run --spec ${CYPRES_SPEC_PATTERN}` so it gets mapped correctly. The e2e config spells it the same so I just kept it the same as that.

keiranjprice101: Cool, just double checking 👍

.github/workflows/test-build.yml
```diff
@@ -329,6 +329,37 @@ jobs:
...
+
+          docker-compose --project-directory $PWD \
+            -f "$REPO_DIR_NAME/docker-compose-stfc.e2e.yml" \
+            up -d mockServer
+
+            sleep 10
```
jekabs-karklins, line 358: We know historically sleeps cause false positive failures. I wonder if it would not be possible to have the same approach as with wait-for-frontend.sh. Check if you could add similar entry point in docker-compose-stfc.e2e.yml like wait-for-mockup-backend.sh

jekabs-karklins (approved):

TickleThePanda (approved):

PR #991, opened 2022-08-16
refactor: Make publication title field multiline

voduguwa:
## Description

Make the publication title field multiline to accommodate for longer titles.

## Motivation and Context

The field accepts 500 characters, so increasing the character limit was not necessary. Longer titles can fit into the text field by being typed on multiple lines.

## How Has This Been Tested
Ensured that you could type the title on multiple lines.

## Fixes

Part of UserOfficeProject/user-office-project-issue-tracker#239


simonfernandes (changes requested):
image

If we allow that many characters, I think it would be good to truncate the display of the title too. Maybe past a certain number of characters it can append ellipses?

voduguwa:
Git issues causing build to fail, so another request has been opened: #1003

PR #999, opened 2022-08-26
Improve GraphQL error handling

mr-field:
## Description
The main point of this PR is to improve how errors during GraphQL calls are handled by the frontend. The main changes are checking for an empty response before trying to handle the error and ensuring that errors are logged to both the client and the server where appropriate.

Another notable change is that users will now be logged out only if the error was caused by an expired session. In all other cases, they should stay logged in.

See the commit descriptions for more details.

## Motivation and Context
Makes it easier to diagnose GraphQL errors through better logging, especially when no response was received.

## How Has This Been Tested
It's very hard to test this as it requires an actual error to happen, but taking the the backend locally resulted in the right error being written to the browser console without the user being logged out. Waiting for thetoken to become invalid both logged the right error and logged out the user.

## Fixes
Closes https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/688
Makes it easier to diagnose problems from https://github.com/UserOfficeProject/user-office-project-issue-tracker/issues/18


TickleThePanda (approved):

TickleThePanda:
Good work. 😄

jekabs-karklins (approved):

PR #1003, opened 2022-09-08
refactor: make title input multiline, truncate longer titles and fix/…

voduguwa:
…add tests

## Description

Make the publication title field multiline to accommodate for longer titles.
Titles that contain more than 30 characters are truncated, so that it fits within the dialogue box. 

## Motivation and Context

- The field accepts 500 characters, so increasing the character limit was not necessary. Longer titles can fit into the text field by being typed on multiple lines.
- Additionally, Titles that are very long do not fit inside of the dialogue box properly.
- Some tests needed to be fixed as changes made to ensure titles could be typed on multiple lines broke some tests. 

## How Has This Been Tested

- Ensured that the user can type on multiple lines through manual testing.
- Added check to test to ensure that long titles are truncated after entry.

## Fixes
Part of UserOfficeProject/user-office-project-issue-tracker#239

### DIFF
--- a/cypress/integration/generic_templates.ts
+++ b/cypress/integration/generic_templates.ts
@@ -346,18 +346,18 @@ context('GenericTemplates tests', () => {
 
       cy.contains('This is a required field');
 
-      const longTitle = faker.lorem.paragraph(5);
+      const longText = faker.lorem.paragraph(5);
 
       cy.get('[data-cy=title-input] textarea')
         .first()
         .clear()
-        .type(longTitle)
-        .should('have.value', longTitle)
+        .type(longText)
+        .should('have.value', longText)
         .blur();
 
       cy.get(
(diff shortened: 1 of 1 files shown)


cypress/integration/generic_templates.ts
```diff
@@ -314,7 +314,7 @@ context('GenericTemplates tests', () => {
       cy.contains(genericTemplateQuestions[1]).should('exist');
     });
 
-    it('Should be able to create proposal with genericTemplate', () => {
+    it.only('Should be able to create proposal with genericTemplate', () => {
```
simonfernandes, line 317: You accidentally left the `only` in here 🙂

simonfernandes (changes requested):

cypress/integration/generic_templates.ts
voduguwa: Resolved! Thank you :)

simonfernandes (approved):

